PR #1, opened 2023-01-25
feat(wallet): add input component

raducristianpopa:
- Add a basic `Input` component
- Simple `Label` component

### DIFF
--- a/packages/wallet/src/ui/Label.tsx
+++ b/packages/wallet/src/ui/Label.tsx
@@ -1,0 +1,20 @@
+import type { ComponentProps, ReactNode } from 'react'
+
+type LabelProps = Omit<
+  ComponentProps<'label'>,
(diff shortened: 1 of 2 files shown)


packages/wallet/src/ui/Label.tsx
```diff
@@ -0,0 +1,20 @@
+import type { ComponentProps, ReactNode } from 'react'
+
+type LabelProps = Omit<
```
Tymmmy, line 3: just a little consistency something here :D , if the previous Input.tsc had Omit details in one row, then this one could have it the same, or Input.tsx can have it in multiple lines as well. Whicever suits you best.

raducristianpopa: This is how Prettier formats at the moment.

Tymmmy (approved):

PR #2, opened 2023-01-25
feat(wallet): add button/link component

raducristianpopa:
- `ButtonOrLink` helper component
- `Button` and `Link` component that are using `ButtonOrLink`

`ButtonOrLink` component renders a `button` or an `a`, based on the props that are passed to it.


Tymmmy (approved):
Nice first start on some default elements we need. And also cool features with the ButtonOrLink component.

PR #39, opened 2023-02-02
feat(wallet): add responsive menu

raducristianpopa:
Adds responsive menu (added a simple icon to every item in the menu).

<details>
<summary>Desktop</summary>
image
</details>

<details>
<summary>Mobile</summary>
image
</details>


packages/wallet/src/components/Menu.tsx
```diff
@@ -0,0 +1,128 @@
...
+  },
+  {
+    name: 'Settings',
+    href: '/settings',
+    icon: <Home className="h-6 w-6" />
+  }
+]
+
+export const Menu = () => {
+  const { pathname } = useRouter()
```
Tymmmy, line 46: pathname should be pathName

packages/wallet/src/components/Menu.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  return (
+    <Disclosure as="nav">
+      {({ open }) => (
+        <>
+          <div className="fixed inset-x-0 flex h-20 flex-col md:inset-y-0 md:h-auto md:w-60">
```
Tymmmy, line 52: the 2 divs act like containers for the menus, both are flex layouts, could they be merged?

packages/wallet/src/components/Menu.tsx
raducristianpopa: This is what `useRouter` is returning.

https://nextjs.org/docs/api-reference/next/router#router-object

packages/wallet/src/components/Menu.tsx
raducristianpopa: The first `div` acts as a wrapper, but we can use the `nav` element as the wrapper and remove the extra `div`.

Tymmmy (approved):

PR #48, opened 2023-02-06
feat(backend): Auth

beniaminmunteanu:
### **PR Summary:**

This PR introduces a new implementation of the authentication system. A previous version that I provided was using redis to store refresh tokens. I now switched to using the postgres db entirely.
The PR targets security, logging, and error handling aspects

### **Implementation Details:**

- The authentication system now uses JSON Web Tokens (JWT) for both access and refresh tokens. The access token is short-lived and is used for authentication, while the refresh token is used for generating new access tokens when they expire.

- The refresh tokens have a default expiration time of 30 days, but this value can be changed in the configuration file.

- The refresh tokens are stored in a separate collection in the database, and the relationship between refresh tokens and users is one-to-many to support multiple devices being logged in by the same user in the future.
(however, currently if a user logs in and already has a refresh token in the database, that one will pe updated, and not a new one created -> which means a user can currently be logged in from a single device)

- The implementation includes the new Winston logger, which provides better logging capabilities and includes the service name, log level, and stack trace in case of errors.

- The implementation also includes a global error handler that catches and logs any unhandled errors that occur in the application.

- The implementation includes a custom Passport JWT strategy that can be applied as a middleware to protect any routes in the future.

### **Authentication Flow:**

1. A user logs in by sending a POST request to the login endpoint with their email and password.
2. If the email and password are valid, the server creates an access token and a refresh token for the user and sends them back to the client via cookies.
3. The client stores the refresh token as an HttpOnly cookie and sends the access token in the Authorization header of subsequent requests.
4. When the access token expires, the client sends a POST request to the refresh endpoint with the refresh token in the cookie.The server verifies the refresh token and generates a new access token if it's valid.
5. If the refresh token is expired or invalid, the server sends a 401 Unauthorized response and the client needs to log in again.


beniaminmunteanu:
Looking for some early feedback on the auth feature, since it starts to have some shape.
As discussed with Radu Popa and Rico Gonzalez, I went for storing the refresh tokens in Redis. Added a docker compose entry.
Running migrations is not yet an automated process.
 
**Login**:
the login process is using the jsonwebtoken package in order to generate accessTokens and refreshTokens, which are then set as cookies on the response.
 
The reason I chose to keep the tokens transfer inside of cookies as opposed to just plain values in a JSON response is that it removes logic from the clients that are using a browser, since the cookies are http only and managed by the browser directly.
 
Access token expiry - 10 minutes
Refresh token expiry - 30 days
(both are customisable as they are being set through environment variables)
(the secrets used in signing the Access tokens and Refresh tokens are environment variables as well)
 
**Signup**:
Just saves the user in DB, no cookie is returned 
(unless you want people to be automatically logged in when their user is successfully created)
 
**Refresh**:
Upon access token expiry, the client will hit this endpoint providing his Refresh token, in order to regenerate a new Access Token (and a new Refresh Token, thus implementing "infinite-login")
 
The Refresh token is checked against the server signature for validity
 
The userId is extracted from the Refresh Token, and it can be used to query against the database in case we need to support features such as "invalidated user"
For the moment, no such DB hit is required.
 
a new Refresh token is created and the old one stored in redis is updated.
 
Both the new Access token and a new Refresh token are sent as cookies back to the client.
 
 
**Passport jwt strategy**
The passport jwt strategy defines how Access cookies are parsed in all the incoming client requests.
It includes validation all these Access tokens from a signature and expiry point of view.
It also adds the decoded jwt information (just the userId) from now to the request object, in case some future routes need to have fast synchronous access to the id of user making the request
 
applying the strategy as a guard to desired protected routes is as simple as:

`mainRouter.post(
  '/protected',
  passport.authenticate('jwt', { session: false }),
  (req, res) => {
    console.log('whatever is here is already authenticated')
  }`

beniaminmunteanu:
After consulting with Alex as well, I will be rewriting the refresh tokens code and give up the redis instance. We would only be using redis if it will also be used for access tokens provided by rafiki. But since we are not sure about how those will be handled, we will go for keeping everything in the main db, favouring flexibility.
This goal is to have a single storage point of all necessary tokens, regardless of the token origin, being it our own auth or rafiki's auth system.

packages/backend/seeds/seed_users.js
```diff
@@ -0,0 +1,70 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.seed = async function (knex) {
+  // Deletes ALL existing entries
+  await knex('users').del()
+  await knex('users').insert([
+    {
+      id: 1,
```
raducristianpopa, line 10: This will have to be an UUID. You can use [faker](https://fakerjs.dev/) to generate data if needed and [uuid](https://www.npmjs.com/package/uuid) for ID's.

packages/backend/tsconfig.json
```diff
@@ -6,6 +6,6 @@
     "rootDir": "./src",
     "declaration": false
   },
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "seeds"],
```
raducristianpopa, line 9: I don't think we need to include `seeds`.

packages/backend/src/index.ts
```diff
@@ -1,25 +1,67 @@
...
+const envSchema = object({
+  PORT: string(),
+  DB_NAME: string(),
+  DB_USERNAME: string(),
+  DB_PASSWORD: string(),
+  JWT_ACCESS_TOKEN_SECRET: string(),
+  JWT_ACCESS_TOKEN_EXPIRATION_TIME: string(),
+  JWT_REFRESH_TOKEN_SECRET: string(),
+  JWT_REFRESH_TOKEN_EXPIRATION_TIME: string()
+})
```
raducristianpopa, line 19: I'll probably move the environment variables to a separate/config file. I don't think we need `dotenv` if we are going to pass the environment variables through Docker.

packages/backend/src/index.ts
```diff
@@ -1,25 +1,67 @@
...
+
+const envPath = resolve(__dirname, '../.env')
+
+const envVars = dotenv.config({ path: envPath })
+
+const envSchema = object({
+  PORT: string(),
+  DB_NAME: string(),
+  DB_USERNAME: string(),
+  DB_PASSWORD: string(),
```
raducristianpopa, line 14: We can use the database URL.

packages/backend/src/index.ts
```diff
@@ -1,25 +1,67 @@
...
 const app: Application = express()
 
 app.disable('X-Powered-By')
 
 app.use(cors())
 app.use(express.json())
+app.use(express.urlencoded({ extended: true }))
+app.use(cookieParser())
 
 const knex = Knex(config[process.env.NODE_ENV || 'development'])
```
raducristianpopa, line 54: Like mentioned above, we can use the database URL. We will probably need to specify the `migrations` folder too

packages/backend/src/errorHandler.ts
```diff
@@ -0,0 +1,12 @@
+import { Request, Response, NextFunction } from 'express'
```
raducristianpopa, line 1: What do you think if we create a `middlewares` folder where we would have the global middlewares?

packages/backend/src/index.ts
raducristianpopa: We can have a `config` folder where we would have `env.ts` and `logger.ts`.

packages/backend/Dockerfile.dev
```diff
@@ -4,7 +4,7 @@ FROM node:lts-hydrogen
 WORKDIR /home/testnet
 
 # Install PNPM
-RUN curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm@7.14.0
+RUN curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm@7.27.0
```
raducristianpopa, line 7: 👍

packages/backend/package.json
```diff
@@ -8,22 +8,26 @@
     "@types/cors": "^2.8.13",
     "bcrypt": "^5.1.0",
     "bcryptjs": "^2.4.3",
```
raducristianpopa, line 10: Do we need both `bcrypt` and `bcryptjs` packages?

packages/backend/package.json
```diff
@@ -8,22 +8,26 @@
...
-    "ioredis": "^5.3.0",
+    "jsonwebtoken": "^9.0.0",
     "knex": "^2.4.2",
     "objection": "^3.0.1",
     "passport": "^0.6.0",
     "passport-jwt": "^4.0.1",
     "pg": "^8.9.0",
     "uuid": "^9.0.0",
-    "validator": "^13.7.0"
+    "validator": "^13.7.0",
```
raducristianpopa, line 22: 🤔 I think we can remove this? Let me know if we will need this later on.

packages/backend/src/auth/auth.service.ts
```diff
@@ -0,0 +1,163 @@
...
+    log.error(error)
+    return res.status(500).json({ error: 'Unable to create user' })
+  }
+}
+
+export const login = async (req: express.Request, res: express.Response) => {
+  try {
+    const {
+      body: { email, password }
+    } = req
```
raducristianpopa, line 61: Let's validate the incoming request's body before using it to query data.

packages/backend/package.json
beniaminmunteanu: no, we don't.

packages/backend/src/index.ts
beniaminmunteanu: moved for now, but maybe we will have a talk about our environement variables.
Also, since knex.js is also a config, we might go back to also moving that into the new config folder, and referencing the file from package.json

packages/backend/src/errorHandler.ts
beniaminmunteanu: mhm

packages/backend/tsconfig.json
beniaminmunteanu: removed

packages/backend/seeds/seed_users.js
beniaminmunteanu: Done, but still have some issues with seeds. Will not use it for now

packages/backend/migrations/20230214221007_refresh_token.js
```diff
@@ -0,0 +1,18 @@
+exports.up = function (knex) {
+  return knex.schema.createTable('refresh_tokens', function (table) {
+    table.increments('id').primary()
```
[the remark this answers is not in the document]
beniaminmunteanu: done

packages/backend/src/index.ts
beniaminmunteanu: I removed the 3 entries and replaced them with a DB_URL env variable:

postgres://postgres:password@postgres/testnet

packages/backend/src/auth/auth.service.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    const user = await User.query()
+      .findOne({ email })
+      .select('id', 'email', 'password')
+      .throwIfNotFound()
```
dragosp1011, line 66: throw an `Invalid credentials` error in this case as well

beniaminmunteanu: It would have been caught by the last catch that throws an invalid credentials with 401 either way.

Also, this is outdated, and the new implementation has a more broad error handling aproach.
Please give it your review

beniaminmunteanu:
added custom errors that all extend BaseError.

The errors are sent "as is" if they reach the global error handler.

Errors that are not instances of BaseError just get logged and 500 Internal Server error is sent.

Added Zod schemas for login and signup.
Added a validator middleware that receives the schema that it needs to validate.

packages/backend/src/config/logger.ts
```diff
@@ -0,0 +1,44 @@
+import { createLogger, format, transports } from 'winston'
```
domin191013, line 1: Well, maybe logger.ts should be placed in another folder like utils, or shared?

beniaminmunteanu: Yeah, it could indeed. It is however a config for the winston instance that we will use. It defines how it should rotate files, where it should store them, etc.

Will see if I come up with a better option. If not, I'd go for moving it inside an utils directory, sure.

docker/dev/docker-compose.yml
```diff
@@ -22,10 +28,14 @@ services:
       - postgres
     environment:
       NODE_ENV: development
-      PORT: 3003
+      PORT: 3000
+    # env_file:
+    #   - ../../packages/backend/.env
```
raducristianpopa, line 33:
```suggestion
      PORT: 3000
      NODE_ENV: development
      DB_URL: postgres://postgres:password@postgres/testnet
      JWT_ACCESS_TOKEN_SECRET: MY_SUPER_SECRET_ACCESS_TOKEN
      JWT_ACCESS_TOKEN_EXPIRATION_TIME: 100000000
      JWT_REFRESH_TOKEN_SECRET: MY_SUPER_SECRET_REFRESH_TOKEN
      JWT_REFRESH_TOKEN_EXPIRATION_TIME: 100000000
```

packages/backend/src/config/envParser.ts
```diff
@@ -0,0 +1,30 @@
...
+    throw new Error('Could not parse environment variables')
+  }
+
+  try {
+    envSchema.parse(envVars.parsed)
+  } catch (err) {
+    console.error('Error parsing environment variables:', err)
+    process.exit(1)
+  }
+}
```
raducristianpopa, line 30:
```suggestion
import { z } from 'zod'

const envSchema = z.object({
  PORT: z.coerce.number(),
  NODE_ENV: z.string().default('DEVELOPMENT'),
  DB_URL: z.string(),
  JWT_ACCESS_TOKEN_SECRET: z.string(),
  JWT_ACCESS_TOKEN_EXPIRATION_TIME: z.coerce.string(),
  JWT_REFRESH_TOKEN_SECRET: z.string(),
  JWT_REFRESH_TOKEN_EXPIRATION_TIME: z.coerce.number()
})


export const parseEnv = () => {
  const result = envSchema.safeParse(process.env)

  if (!result.success) {
    console.error(
      'Error parsing environment variables:',
      result.error.flatten()
    )
    process.exit(1)
  }

  return result.data
}

export const env = parseEnv()

```

raducristianpopa: We can also rename the file to `env.ts`.

packages/backend/src/config/logger.ts
```diff
@@ -0,0 +1,44 @@
...
+import DailyRotateFile from 'winston-daily-rotate-file'
+import path from 'path'
+import { existsSync, mkdirSync } from 'fs'
+
+const env = process.env.NODE_ENV || 'development'
```
raducristianpopa, line 6: With the changes mentioned above we should be able to do:

```suggestion
import { existsSync, mkdirSync } from 'fs'
import { env } from './env'
```

packages/backend/src/config/logger.ts
```diff
@@ -0,0 +1,44 @@
...
+      format.errors({ stack: true }),
+      format.printf(
+        (info) =>
+          `[${info.timestamp}] [${info.level}] [${name}] ${info.message}` +
+          (info.stack ? `\n${info.stack}` : '')
+      )
+    ),
+    transports: [
+      new transports.Console({
+        level: env === 'development' ? 'debug' : 'info'
```
raducristianpopa, line 38:
```suggestion
        level: env.NODE_ENV === 'development' ? 'debug' : 'info'
```

packages/backend/src/middlewares/validator.ts
```diff
@@ -0,0 +1,18 @@
...
+  return (req: Request, _res: Response, next: NextFunction) => {
+    try {
+      schema.parse(req.body)
+      next()
+    } catch (error) {
+      if (error instanceof ZodError) {
+        throw new BadRequestException(error.errors[0].message)
+      }
+      next(error)
+    }
```
raducristianpopa, line 16: We can avoid using `parse` since we will have to wrap everything in a `try-catch` block. I can suggest using `safeParse`:

```suggestion
    const result = schema.safeParse(req.body)
    
    if(!result.success) {
      throw new BadRequestException(error.errors[0].message)
    }
    
    next()
```

raducristianpopa: 🤔 I think by having a middleware that is validating a schema we would not get the types for the request's body in a controller.

packages/backend/src/index.ts
```diff
@@ -1,25 +1,39 @@
...
 
 app.use(passport.initialize())
 
+passport.use(jwtStrategy)
+
+app.use(mainRouter)
+
+app.use(errorHandler)
+
 app.listen(process.env.PORT, (): void => {
```
raducristianpopa, line 37: If we are going with the changes mentioned above, we should import the `env` object and change this to:
```suggestion
app.listen(env.PORT, (): void => {
```

raducristianpopa: Ditto for every `process.env.VARIABLE`.

packages/backend/src/auth/schemas/loginSchema.ts
```diff
@@ -0,0 +1,3 @@
+import { signupSchema } from './signupSchema'
+
+export const loginSchema = signupSchema
```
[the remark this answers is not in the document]
dragosp1011: IMO having everything separated gives more visibility and with IDEs you can quickly access them anyway

packages/backend/src/auth/schemas/signupSchema.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from 'zod'
+
+export const signupSchema = z.object({
```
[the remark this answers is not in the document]
dragosp1011: I think for signup there should be a little bit more details

packages/backend/src/middlewares/validator.ts
beniaminmunteanu: removed the validator and went for a different approach which is able to infer types. 
The validation will be done at the controller level, no longer a middleware that precedes the controller.

packages/backend/src/auth/schemas/loginSchema.ts
raducristianpopa, line 3:
```suggestion
export const logInSchema = z.object({
  email: z.string().email(),
  password: z.string()
})
```

packages/backend/src/auth/schemas/signupSchema.ts
```diff
@@ -0,0 +1,6 @@
...
+  email: z.string().email(),
+  password: z.string()
+})
```
raducristianpopa, line 6:
```suggestion
import { z } from 'zod'
import { loginSchema } './loginSchema'

export const signupSchema = loginSchema.merge(z.object({
  confirmPassword: z.string()
}))
```

A password confirmation is needed for registration:
https://github.com/interledger/testnet/blob/08b8ed175f0c6b5567bc6b0462f1d01d01b8577d/packages/wallet/src/pages/auth/register.tsx#L10-L26

packages/backend/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "lib": ["ESNext"],
     "outDir": "./dist",
     "rootDir": "./src",
-    "declaration": false
+    "declaration": false,
+    "sourceMap": true
```
raducristianpopa, line 8: Why do we need the source maps? 🤔

packages/backend/src/middlewares/validator.ts
```diff
@@ -0,0 +1,18 @@
+// import { HTTP400Error } from '../utils/httpErrors'
```
raducristianpopa, line 1: Not a middleware now, we can move it outside of this folder.
A good enhancement would be if we validate the whole request object (including `params` and `queryParams`). Can keep things as is for now and incrementally adopt new changes.

packages/backend/tsconfig.json
beniaminmunteanu: for debugging

packages/backend/src/auth/schemas/signupSchema.ts
beniaminmunteanu: Zod is cool!

docker/dev/docker-compose.yml
```diff
@@ -8,10 +12,12 @@ services:
       POSTGRES_PASSWORD: password
       POSTGRES_DB: testnet
     ports:
-      - '5433:5432'
+      - '5432:5432'
```
dragosp1011, line 15: 5432 will conflict with rafiki local setup, dsmr fot 3000

beniaminmunteanu: So should we let rafiki use the defaults and we change testnet itself to 5433 and idk 3003?

packages/backend/src/auth/auth.service.ts
```diff
@@ -0,0 +1,152 @@
...
+      expiresIn: env.JWT_REFRESH_TOKEN_EXPIRATION_TIME
+    }),
+    expiresIn: Number(env.JWT_REFRESH_TOKEN_EXPIRATION_TIME)
+  }
+}
+
+export const signup = async (req: express.Request, res: express.Response) => {
+  try {
+    const { email, password, confirmPassword } = await zParse(signupSchema, req)
+    if (password !== confirmPassword) {
```
dragosp1011, line 39: is it possible to have this check in validation?

packages/backend/src/auth/auth.service.ts
```diff
@@ -0,0 +1,152 @@
...
+      generateRefreshToken(userId)
+
+    existingRefreshToken.token = newRefreshToken
+    existingRefreshToken.userId = user.id
+    existingRefreshToken.expiresAt = new Date(
+      Date.now() + refreshTokenExpiresIn * 1000
+    )
+    await existingRefreshToken.$query().patch(existingRefreshToken)
+
+    res.cookie('AccessToken', newAccessToken, {
```
dragosp1011, line 137: looks similar to what happens at login, maybe extract in a method

packages/backend/src/errors/unauthorisedException.ts
```diff
@@ -0,0 +1,7 @@
+import { BaseError } from './baseError'
+export class UnauthorisedException extends BaseError {
```
dragosp1011, line 2: I think it will be better to have errors/exceptions in each module, and here only some generic once
also if the file is a class I'll propose to have the class name as file name (starting with capital)

dragosp1011 (dismissed):

packages/backend/src/auth/auth.service.ts
beniaminmunteanu: I tried to do it as Radu suggested, with the Zod object chained with a .superRefine, as in this screenshot.

But this modifies the return type of the Zod object, and it is no longer compatible with the z.infer that we use to keep the types returned by the zParse.

That is why I moved it just as a validation in the controller for now.

packages/backend/src/errors/unauthorisedException.ts
beniaminmunteanu: Hmm, I agree. I only kept it all in a single common error directory because there were only 2 extensions of the BaseError for now.

I will move the Unauthorised exception in the auth module.
BadRequest should be for every module, and as such I will keep it here along with the BaseError

dragosp1011 (approved):

PR #56, opened 2023-02-09
feat(wallet): Create New Wallet User Account

### LINKED ISSUE #18: Create New Wallet Account/Register functionality (FE)
Create New Wallet Account/Register functionality for testnet Wallet
image

Tymmmy:
### Context
- fixes #18

### Changes
Added form to create new Wallet User Account


packages/wallet/src/ui/Form.tsx
```diff
@@ -63,7 +63,7 @@ export const Form = <T extends FieldValues>({
 }: FormProps<T>) => {
   return (
     <FormProvider {...form}>
-      <form onSubmit={form.handleSubmit(onSubmit)} {...props}>
+      <form onSubmit={form.handleSubmit(onSubmit)} {...props} className="w-3/5">
```
raducristianpopa, line 66: Let's not apply styles to the form abstraction. We can just create a wrapper around the `Form` component and it will have it's parent width.

packages/wallet/src/pages/register.tsx
```diff
@@ -0,0 +1,85 @@
...
+        />
+        <button
+          aria-label="login"
+          type="submit"
+          className="m-auto py-2 sm:py-5"
+        >
+          <Play />
+        </button>
+      </Form>
+      <div className="absolute bottom-0 h-[200px] w-full bg-[url('../../public/leafs.svg')] bg-contain bg-center bg-no-repeat md:hidden"></div>
```
raducristianpopa, line 74: Should we create a component instead of passing it as a background (similar to how `Bubbles` is displayed in the application) ? 🤔

Tymmmy: It's not really used anywhere else. Don't think a component is necesary.

Tymmmy: I added it a background so the text will be in the div also, and this way the position of text and image is handled. They need to overlap.

raducristianpopa (approved):

PR #104, opened 2023-02-20
feat(rafiki): add rafiki as a submodule and setup local instance

dragosp1011:
### Changes
- add rafiki as a submodule
- add rafiki services to dev docker compose


docker/dev/docker-compose.yml
```diff
@@ -10,24 +10,30 @@ services:
     ports:
       - '5433:5432'
     restart: unless-stopped
+    networks:
+      - testnet
     volumes:
       - pg-data:/var/lib/postgresql/data
+      - ./dbinit.sql:/docker-entrypoint-initdb.d/init.sql
```
raducristianpopa, line 17: Should we have two `postgres` images (one for the wallet and one for Rafiki)?

docker/dev/docker-compose.yml
```diff
@@ -46,5 +54,117 @@ services:
...
+      dockerfile: ./packages/http-signature-utils/Dockerfile
+    restart: always
+    ports:
+      - '3042:3000'
+    environment:
+      KEY_FILE: /workspace/private-key.pem
+    volumes:
+      - ./private-key.pem:/workspace/private-key.pem
+    networks:
+      - testnet
```
raducristianpopa, line 158: Do we need the signatures application? 🤔

https://github.com/interledger/rafiki/tree/main/packages/http-signature-utils#http-signature-utils-library

docker/dev/docker-compose.yml
dragosp1011: for dev I don't think we need to separate them

docker/dev/docker-compose.yml
dragosp1011: looks like no, we will add it back if we will need it in the future

docker/dev/dbinit.sql
```diff
@@ -0,0 +1,7 @@
+CREATE USER backend WITH PASSWORD 'backend';
+CREATE DATABASE backend;
+ALTER DATABASE backend OWNER TO backend;
+
+CREATE USER auth WITH PASSWORD 'auth';
+CREATE DATABASE auth;
+ALTER DATABASE auth OWNER TO auth;
```
beniaminmunteanu, line 7: These are both used in rafiki?
Or is the backend you are refering to there our testnet backend?

dragosp1011: both are for rafiki services

beniaminmunteanu: got it

beniaminmunteanu (dismissed):
LGTM

raducristianpopa (approved):

PR #107, opened 2023-02-21
feat(wallet): add account page

### LINKED ISSUE #35: Create list of Payment Pointers page (FE)
Create page for listing payment pointers for a selected account

image

image

raducristianpopa:
### Context
- closes #35

### Changes


packages/wallet/src/components/PaymentPointerCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+              View
+            </PaymentPointerCardButton>
+          </>
+        )}
+      </div>
+    </div>
+  )
+}
+
+const cardStyles = `
```
raducristianpopa, line 159: Placed the card styles here because prettier was rearranging them, if they were placed inside the component.

packages/wallet/src/pages/account/[id].tsx
```diff
@@ -0,0 +1,141 @@
...
+              Exchange
+            </span>
+          </Link>
+        </div>
+        <div className="flex items-center justify-between">
+          <h3 className="text-lg font-semibold leading-none text-green">
+            Account
+          </h3>
+        </div>
+        <div className="flex items-center justify-between rounded-md bg-gradient-to-r from-[#92DBCA] to-[#56B1AF] px-3 py-2">
```
Comment (Tymmmy) [inline packages/wallet/src/pages/account/[id].tsx:70]: couldn't we use a gradient we already have in the config?

Comment (raducristianpopa) [reply packages/wallet/src/pages/account/[id].tsx:70]: Totally missed that. Good catch.

Tymmmy (approved):

Issue #5, opened 2023-02-02
Create Login and New Account/Register functionality


Tymmmy:
#Acceptance Criteria

- [x] #16 
- [x] #6 
- [x] #15 
- [x] #17 
- [x] #18 
- [x] #134
- [x] #136

#Design
![image](https://user-images.githubusercontent.com/117268143/217319057-a220fbb1-d433-4776-ad2b-96e2d247af84.png)

<img width="627" alt="image" src="https://user-images.githubusercontent.com/117268143/216281313-05cecd01-d5b4-4e56-b8c6-05131ba85df0.png">
<img width="623" alt="image" src="https://user-images.githubusercontent.com/117268143/216281378-97b0576a-963e-4a17-b617-49e35925bcd0.png">

Tymmmy:
Main subtasks are implemented for this Epic story.
Other bugfixes will probably be standalone tickets.

PR #132, opened 2023-02-27
fix: #5 auth

beniaminmunteanu:
Auth service properly throws errors
Response type defined
fix user foreign key reference typo

### Context
- fixes #
The user model reference to refresh tokens table had a typo in it.
The docker compose did not map ports properly for postgres.
Controller Errors were not properly thrown (and caught) by the general error handler.

### Changes
Added a Response type that should be used for all testnet endpoints
The validator now defines an error object that provides the path and issue of each field sent as input.
Since we are using express <5, the next(err) method must be called specifically. Controllers are now surrounded by try catches, and in the catch block we "next" the errors to be caught by the custom framework-scoped error handler

Example of 400 Bad request error received now:
image


packages/backend/src/auth/auth.service.ts
```diff
@@ -72,58 +73,74 @@ export const signup = async (req: express.Request, res: express.Response) => {
...
+  res: express.Response,
+  next: NextFunction
+) => {
+  try {
+    const { email, password } = await zParse(loginSchema, req)
 
-  const user = await User.query()
-    .findOne({ email })
-    .select('id', 'email', 'password')
-    .throwIfNotFound(new UnauthorisedException('Invalid credentials'))
```
beniaminmunteanu, line 81: This did not work. It throws a NotFound Error, not an Unauthorised Exception. The object passed here to the throwIfNotFound method only adds it to the "data" property of the NotFoundError. It does not replace the instance with an UnauthroisedException one.

The only way to change this behaviour is by doing it globally on the knex Model instance in index.ts. We can define a custom class there to be thrown instead of NotFound. But I'd rather just not use the feature at all, and just check if the user exists or not as I did in this change request

packages/backend/src/middlewares/validator.ts
```diff
@@ -1,16 +1,25 @@
...
-      throw new BadRequestException(error.errors[0].message)
+      const errors = error?.errors?.reduce((acc, error) => {
+        const newAcc = {
+          ...acc,
+          [error.path[0]]: error.message
+        }
+        return newAcc
+      }, {})
+
+      throw new BadRequestException('Invalid input', errors)
```
raducristianpopa, line 22: You can use `flatten` for this (https://github.com/colinhacks/zod/blob/master/ERROR_HANDLING.md#flattening-errors)

packages/backend/src/middlewares/errorHandler.ts
```diff
@@ -9,8 +9,10 @@ export const errorHandler = (
...
   if (err instanceof BaseError) {
-    const { statusCode, message } = err
-    res.status(statusCode).json({ error: message })
+    res
+      .status(err.statusCode)
+      .json({ success: err.success, message: err.message, errors: err.errors })
+    throw err
   } else {
     log.error(err)
     res.status(500).json({ error: 'Internal Server Error' })
```
raducristianpopa, line 18:
```suggestion
    res
      .status(err.statusCode)
      .json({ success: err.success, message: err.message, errors: err.errors })
    throw err
  } else {
    log.error(err)
    res.status(500).json({ success: false, message: 'Internal Server Error' })
```

packages/backend/src/auth/auth.service.ts
```diff
@@ -72,58 +73,74 @@ export const signup = async (req: express.Request, res: express.Response) => {
...
+    appendTokensToCookie(
+      res,
+      accessToken,
+      accessTokenExpiresIn,
+      refreshToken.token,
+      refreshTokenExpiresIn
+    )
 
-  return res.send({ user })
+    return res.send({ user })
```
raducristianpopa, line 132: This should be returned as a `Response`:
https://github.com/interledger/testnet/blob/38f401b207a0bd8f6778bc65b5fbd50883310fc2/packages/backend/src/shared/models/Response.ts#L1

I'll probably rename the interface since there already two `Response` types: from `node` and `express`

packages/backend/src/middlewares/validator.ts
beniaminmunteanu: I used flatten, and also tried giving it a custom mapper as in these docs:
image

However, it did not help much, since I was forced to return an object for each fieldError key, instead of the string that we wanted. 

My latest code maps the arrays to strings.

raducristianpopa (dismissed):

packages/backend/src/auth/auth.service.ts
```diff
@@ -61,79 +62,103 @@ export const signup = async (req: express.Request, res: express.Response) => {
...
+        Date.now() + refreshTokenExpiresIn * 1000
+      )
 
-    await user.$relatedQuery('refreshTokens').insert(refreshToken)
-  }
+      await refreshToken.$query().patch(refreshToken)
+    } else {
+      refreshToken = new RefreshToken()
+      refreshToken.token = generatedToken
+      refreshToken.expiresAt = new Date(
```
dragosp1011, line 117: us a method that transforms `expiresIn` to `expiresAt`
or maybe one that gets the refresh token and applies the new values (token + expiresAt)

dragosp1011 (dismissed):

packages/backend/src/auth/auth.service.ts
beniaminmunteanu: Good point. 
I used the RefreshToken constructor for this purpose and modified the code in auth service to use it accordingly.

dragosp1011 (approved):

dragosp1011 (approved):

PR #144, opened 2023-03-02
feat(wallet): Implement requests for wallet creation and verify identity

### LINKED ISSUE #51: Implement KYC - send personal data - verify identity (FE)
Validate User Account with KYC Wallet information:
image

Send data to /wallet for Rapyd wallet creation (Personal Details tab)
Send data to /verify to Verify Identity in Rapyd (ID Proof tab)

Tymmmy:
### Context
- fixes #51 

### Changes
Added /wallet request to send personal data in order to create Rapyd wallet
Added /verify request to send data for identity verification


packages/wallet/src/components/kyc/context.tsx
```diff
@@ -3,6 +3,8 @@ import { createContext, useContext, useState, type ReactNode } from 'react'
 type KYCFormContextProps = {
   tab: number
   setTab: (tab: number) => void
+  disabled: boolean
+  setDisabled: (isDisabled: boolean) => void
```
raducristianpopa, line 7: The `disabled` state is a general state for the whole KYC or we need this to work per tab?

packages/wallet/src/components/dialogs/SuccessDialog.tsx
```diff
@@ -59,25 +61,28 @@ export const SuccessDialog = ({
...
-                    Close
-                  </Button>
                   {redirect && (
                     <Button
                       intent="secondary"
                       aria-label={redirectText ?? 'redirect'}
                       fullWidth
                       href={redirect ?? '/'}
-                      onClick={() => onClose()}
+                      onClick={() => {
```
raducristianpopa, line 70: I've understood the idea of `onSuccess`. The problem is that we have an `href` and `onClick`. These will both execute when the button is clicked.

packages/wallet/src/components/dialogs/SuccessDialog.tsx
```diff
@@ -59,25 +61,28 @@ export const SuccessDialog = ({
...
                     </Button>
                   )}
+                  <Button
+                    intent="success"
+                    aria-label="close dialog"
+                    fullWidth
+                    onClick={() => onClose()}
+                  >
+                    Close
+                  </Button>
```
raducristianpopa, line 85: 🤔 I don't think we need the close button. Clicking it, it will just close the dialog and not redirect the user to the next tab.

packages/wallet/src/components/kyc/context.tsx
Tymmmy: Currently it is a general state, so that user doesn't navigate to ID Proof or Bank Account before feeling up Personal Details. In the future it might be a tab disabled. But the way it is currently working it is per whole group.

raducristianpopa: Wouldn't this mean that the user is going to be able to go the first tab or the last when we change `disabled` to `false`? 
https://github.com/interledger/testnet/blob/5f050932fb23b5724316d56909507241795d4a4b/packages/wallet/src/components/kyc/PersonalDetailsForm.tsx#L44-L54

packages/wallet/src/components/dialogs/SuccessDialog.tsx
```diff
@@ -59,25 +75,26 @@ export const SuccessDialog = ({
...
                 <div className="mt-5 grid grid-cols-1 gap-3">
-                  <Button
-                    intent="success"
-                    aria-label="close dialog"
-                    fullWidth
-                    onClick={() => onClose()}
-                  >
-                    Close
-                  </Button>
                   {redirect && (
```
raducristianpopa, line 78: Can we remove the `redirect` check now since we are handling this with `successButtonProps`? 🤔

raducristianpopa (approved):

PR #147, opened 2023-03-03
bug(wallet): fix issue with Safari and svg element

### LINKED ISSUE #146: Safari doesn't display edit icons on payment pointer list
image

Tymmmy:
### Context
- fixes #146 

### Changes
Added width and height to svg elements for them to be displayed correctly in Safari as well
Arranged the Payment pointer cards to be the same distance from top to bottom

### DIFF
--- a/packages/wallet/src/components/PaymentPointerCard.tsx
+++ b/packages/wallet/src/components/PaymentPointerCard.tsx
@@ -157,7 +157,7 @@ export const PaymentPointerCard = ({
 }
 
 const cardStyles = `
-flex items-center justify-between border-b border-b-green-4 px-2 py-3
+flex items-center justify-between border-b border-b-green-4 px-2 pb-3 pt-1
 [&:nth-child(4n+1)_div_button]:bg-green-5 [&:nth-child(4n+1)_div_button:hover]:bg-green-6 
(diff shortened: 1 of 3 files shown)


packages/wallet/src/components/PaymentPointerCard.tsx
```diff
@@ -157,7 +157,7 @@ export const PaymentPointerCard = ({
 }
 
 const cardStyles = `
-flex items-center justify-between border-b border-b-green-4 px-2 py-3
+flex items-center justify-between border-b border-b-green-4 px-2 pb-3 pt-1
 [&:nth-child(4n+1)_div_button]:bg-green-5 [&:nth-child(4n+1)_div_button:hover]:bg-green-6
```
Tymmmy, line 161: there is a mt-8 for the payment pointer cards, that I could not overwrite :D  SO I added less padding to top then bottom, so that the cards look equaldistance from toop and bottom

raducristianpopa: `space-y-2` adds some spacing between the cards. We can change this value and keep `py-3` for `PaymentPointerCard`.

https://github.com/interledger/testnet/blob/4432c4ce7b475ce3c5bb4c7f510b3c158aa15942/packages/wallet/src/pages/account/%5Bid%5D.tsx#L78-L85

raducristianpopa (approved):

PR #156, opened 2023-03-06
feat(wallet): bump `react-hook-form` version

raducristianpopa:
### Context
- bumps `react-hook-form` version
- https://github.com/interledger/testnet/pull/145

### Changes
- Moved `useZodForm` inside the `hooks` folder
- Removed `onSubmit` functions and moved them inside the `onSubmit` prop.


packages/wallet/src/ui/forms/Form.tsx
```diff
@@ -26,40 +22,25 @@ import { XCircle } from '@/components/icons/X'
  *      });
  *
  *      return (
- *            <Form form={form} onSubmit={(data) => {
```
Tymmmy, line 29: do we need the comments?

raducristianpopa: This is the general usage example on how to use the `Form` component along with `useZodForm`.

Tymmmy (approved):

PR #170, opened 2023-03-07
feat: create account and payment pointer

### LINKED ISSUE #29: Create New Payment pointer (BE)
Create Payment Pointer for selected account.
Users should be able to select payment pointers (public names) as they are selecting new email addreses.

Documentation for paymentPointers:
https://paymentpointers.org

### LINKED ISSUE #31: Create Add New Currency/Financial Account Page (BE)
Create a new currency/financial/moeny account page.

The data from Frontend will be 👍 
image

### LINKED ISSUE #36: Create Payment Pointers list page (BE)
Return list of payment pointers for the selected account

image

image

dragosp1011:
### Context
- fixes #29 
- fixes #31 
- fixes #36

### Changes
- Add client for rafiki graphql requests
- Create/get/list assets from rafiki
- Create/get/list accounts
- Create rafiki payment pointers
- Get/list payment pointers for account


packages/backend/src/rafiki/request/asset.request.ts
```diff
@@ -0,0 +1,88 @@
+import {Asset, AssetEdge, AssetMutationResponse, AssetsConnection} from "../generated/graphql";
```
raducristianpopa, line 1: Everything inside `src/rafiki/` is not properly formatted because we ignore the `rafiki` folder.

https://github.com/interledger/testnet/blob/022488fe8426ac448192d320e937fcd94748cdd1/.prettierignore#L12

packages/backend/src/rafiki/request/asset.request.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+  `
+    const createAssetInput = {
+        input: {
+            code,
+            scale
+        }
+    }
+
+    return requestGQL<{ createAsset: AssetMutationResponse}>(createAssetMutation, createAssetInput, {})
```
raducristianpopa, line 30: Like mentioned above, we can use `graphl-codegen` to also generate operation types directly ([typescript-operations](https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-operations)).

packages/backend/src/rafiki/graphql.client.ts
```diff
@@ -0,0 +1,14 @@
+import {request} from 'graphql-request'
+import env from '../config/env'
+import {Mutation, Query} from "./generated/graphql";
+
+const GRAPHQL_ENDPOINT = env.GRAPHQL_ENDPOINT
+
+export async function requestGQL<T extends Partial<Mutation | Query>>(query: string, variables?: Record<string, unknown>, headers?: Record<string, string>): Promise<T> {
```
raducristianpopa, line 7: Should we have a default for `headers` that is set to an empty object? This way we will not have to pass an empty object as the 3rd arg when using `requestGQL`.

packages/backend/src/rafiki/request/payment-pointer.request.ts
```diff
@@ -0,0 +1,44 @@
...
+          publicName
+        }
+      }
+    }
+  `
+    const createPaymentPointerInput = {
+        assetId: assetId,
+        url: `${OPEN_PAYMENTS_HOST}/${accountName}/${paymentPointerName}`,
+        publicName: paymentPointerName
+    }
```
raducristianpopa, line 31: The account name is not unique at the moment. We can add a `unique` constraint, but I don't think it's a good idea. The users should be able to define their account name (Alice and Bob can both have an account named `USD Primary`). 

Having a check that verifies if the user already has an account with the given name would be okay (Bob should not be able to create two accounts with the name `USD Secondary`).

I think  we should drop `accountName` when creating a payment pointer and just rely on `paymentPointerName`.

packages/backend/src/routes.ts
```diff
@@ -34,6 +49,57 @@ mainRouter.post(
   }
 )
 
+// asset
+mainRouter.post(
+  '/assets',
+  passport.authenticate('jwt', { session: false }),
+  createAsset
+)
```
raducristianpopa, line 57: In what scenarios we will have to hit this endpoint?

dragosp1011: it is more like an admin endpoint, so you can add some assets for the whole app
maybe make sense to not have it and just use hit rafiki directly for that?

packages/backend/src/rafiki/request/asset.request.ts
dragosp1011: added a separate issue: https://github.com/interledger/testnet/issues/173

packages/backend/src/routes.ts
raducristianpopa: > maybe make sense to not have it

I think so. Since the first scaffold for Rafiki Admin is almost done,(this includes the possibility to create an asset), we should be able create assets directly from there or just seed the database when the docker container spins up.

packages/backend/src/rafiki/graphql.client.ts
dragosp1011: removed empty objects headers

packages/backend/package.json
```diff
@@ -3,7 +3,9 @@
   "scripts": {
     "start": "node dist/index.js",
     "build": "tsc --build tsconfig.json",
-    "knex": "knex"
+    "knex": "knex",
+    "copy:graphql-types": "cp ../../rafiki/packages/backend/src/graphql/generated/graphql.ts ./src/rafiki/generated",
```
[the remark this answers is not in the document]
dragosp1011: created a separate issue: https://github.com/interledger/testnet/issues/173

packages/backend/src/account/errors/AccountExistsException.ts
```diff
@@ -0,0 +1,7 @@
+import { BaseError } from '../../shared/models/errors/BaseError'
+export class AccountExistsException extends BaseError {
+  constructor(name: string) {
+    super(429, `Account with ${name} already exists`)
```
beniaminmunteanu, line 4: 429 is ment for "Too many requests"
I belive you meant 409?

Also, I would Just move this out of the account module.
Let's have it as a generic ConflictException in the shared/errors directory.
You then instantiate it in the account service with the business-related message
new ConflictException('Account with ${name} already exists')

packages/backend/src/account/account.service.ts
```diff
@@ -0,0 +1,73 @@
...
+  try {
+    const { name, assetRafikiId } = await zParse(accountSchema, req)
+    const userId = getUserIdFromRequest(req)
+
+    const existentAccount = await Account.query()
+      .where('userId', userId)
+      .where('name', name)
+      .first()
+    if (existentAccount) {
+      new AccountExistsException(name)
```
beniaminmunteanu, line 24: 👀 -> AccountExistsException.ts Code review suggestion:
https://github.com/interledger/testnet/pull/170/files#r1131295734

packages/backend/src/payment-pointer/payment-pointer.service.ts
```diff
@@ -0,0 +1,95 @@
...
+  next: NextFunction
+) => {
+  try {
+    const userId = getUserIdFromRequest(req)
+    const accountId = req.params.accountId
+    // Validate that account id belongs to current user
+    const account = await Account.query()
+      .findById(accountId)
+      .where('userId', userId)
+      .throwIfNotFound()
```
beniaminmunteanu, line 55: Same here, I suggest checking the existance of the account after the query is being made, and throwing a custom error manually.
Maybe it's BadRequest or Forbidden(403)

```suggestion
      if(!account){
         throw new BadRequestException()
      }
```

packages/backend/src/payment-pointer/payment-pointer.service.ts
```diff
@@ -0,0 +1,95 @@
...
+) => {
+  try {
+    const userId = getUserIdFromRequest(req)
+    const accountId = req.params.accountId
+    const id = req.params.id
+    // Validate that account id belongs to current user
+    await Account.query()
+      .findById(accountId)
+      .where('userId', userId)
+      .throwIfNotFound()
```
beniaminmunteanu, line 84: ^^

packages/backend/src/payment-pointer/payment-pointer.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const paymentPointer = await PaymentPointerModel.query()
+      .findById(id)
+      .where('accountId', accountId)
+      .throwIfNotFound()
```
beniaminmunteanu, line 89: well you get the gist by now, I will just be adding comments on these so you don't miss some :D

packages/backend/src/rafiki/request/asset.request.ts
```diff
@@ -0,0 +1,91 @@
...
+  AssetEdge,
+  AssetMutationResponse,
+  AssetsConnection
+} from '../generated/graphql'
+import { gql } from 'graphql-request'
+import { requestGQL } from '../graphql.client'
+
+export async function createAsset(code: string, scale: number): Promise<Asset> {
+  const createAssetMutation = gql`
+    mutation CreateAsset($input: CreateAssetInput!) {
```
beniaminmunteanu, line 12: do we have any way of generating these somewhere, or are we stuck with using magic strings for now?

packages/backend/src/types/express/index.d.ts
```diff
@@ -0,0 +1,12 @@
+export {}
+
+declare global {
+  namespace Express {
+    interface User {
+      id: string
```
beniaminmunteanu, line 6: I belive this is extending the Request express type to also contain the user,right?
If such, add email aswell, since it is also on the req.user now

beniaminmunteanu, line 6:
```suggestion
      id: string
      email: string
```

packages/backend/src/rafiki/request/asset.request.ts
dragosp1011: I did not find a way but I am new to graphql staff, @raducristianpopa maybe you know if it is possible?

raducristianpopa: There is not really a way to generate them - that's the query language. Queries and mutations can be written in a separate file (for ie: queries can be written in `queries.graphql` and mutations in `mutations.graphql`), but I think it's better to keep them where they are actually used/needed (how it's currently done).

With `graphql-codegen` we can generate the operations (query or mutation) types, so we won't have to create them ourselves and get some type safety along the way. Either way, if we have a typo in a query or a mutation (instead of `CreateAssetInput` it's `CreateAsetInput`) `graphql-codegen` will fail because it would not be able to find the bad written field/input.

beniaminmunteanu: and is graphql-codegen to be integrated in this PR aswell, to provide types? Or should we consider it a follow-up

dragosp1011: added a separate issue for looking into this https://github.com/interledger/testnet/issues/173

beniaminmunteanu (approved):

PR #174, opened 2023-03-08
feat(backend): Added document type api

domin191013:
### Context
Added document type api

### Changes


packages/backend/src/rapyd/documents/documents.service.ts
```diff
@@ -0,0 +1,40 @@
...
+import { makeRapydGetRequest } from '../utills/request'
+
+const log = logger('rapydService')
+
+export const getDocumentTypes = async (
+  req: Request,
+  res: Response<BaseResponse>
+) => {
+  try {
+    const { country } = req.query
```
beniaminmunteanu, line 13: maybe verify the existance of country inside the query object aswell before actually making the http request further down, so you treat the case of sending a country= (nothing) query param to rapyd, which would probably still return something but it would be unexpected

beniaminmunteanu (approved):

PR #177, opened 2023-03-08
feat(backend): /verify and /documents api improvement

domin191013:
### Context
Improved 
/documents api, 
/verify api 

### Changes


packages/backend/src/rapyd/documents/documents.service.ts
```diff
@@ -0,0 +1,48 @@
...
+import { User } from '../../user/models/user'
+
+const log = logger('rapydService')
+
+export const getDocumentTypes = async (
+  req: Request,
+  res: Response<BaseResponse>
+) => {
+  try {
+    const { email } = req.user as User
```
beniaminmunteanu, line 14: req.user also has the id. Maybe use that to query the db for the country retrieval, since the id being the primary key is also indexed, the email is not.

packages/backend/src/wallet/wallet.service.ts
```diff
@@ -76,3 +83,52 @@ export const createWallet = async (
...
+
+    const {
+      document_type,
+      front_side_image,
+      front_side_image_mime_type,
+      face_image,
+      face_image_mime_type
+    } = await zParse(kycSchema, req)
+
+    const user = await User.query().where('email', email).first()
```
beniaminmunteanu, line 102: same here, maybe going for a find by id would be better performance wise because of the index

beniaminmunteanu (approved):

PR #179, opened 2023-03-09
feat(testing): Integration tests

### LINKED ISSUE #121: Auth Unit tests

beniaminmunteanu:
### Context
- closes #121 

This PR has the first functional jest tests in the testne application. 

Mocking Objection models was quite troublesome. We decided to go for integration tests instead, resulting in not mocking the db interaction entirely, **but spinning up a postgres docker container that is used in the tests**.

This results in a higher reliability, as anything can go wrong with an actual container, opposed to having pre-defined monkey wrenches in our flows in the unit tests, with predictable errors and results.

The PR introduces the first 2 tests to serve as examples of use, consisting of a successful and a failed test.


packages/backend/src/config/env.ts
```diff
@@ -1,35 +1,39 @@
...
+//   NODE_ENV: z.string().default('development'),
+//   DB_URL: z.string(),
+//   JWT_ACCESS_TOKEN_SECRET: z.string(),
+//   JWT_ACCESS_TOKEN_EXPIRATION_TIME: z.coerce.string(),
+//   JWT_REFRESH_TOKEN_SECRET: z.string(),
+//   JWT_REFRESH_TOKEN_EXPIRATION_TIME: z.coerce.number(),
+//   RAPYD_API: z.string(),
+//   RAPYD_ACCESS_KEY: z.string(),
+//   RAPYD_SECRET_KEY: z.string()
+// })
```
raducristianpopa, line 17: Instead of commenting out the `envSchema` we can specify defaults for our environment variables:
```
PORT: z.coerce.number().default(3000)
```

raducristianpopa (commented):
I'm getting a lot of type errors. We might want to move the `tests` folder inside `src/` or remove `rootDir` and add the folder to `includes` in `tsconfig.json`.

packages/backend/src/config/env.ts
beniaminmunteanu: indeed

dragosp1011: can we set some test values on jest setup?

packages/backend/tests/auth/auth.service.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+describe('AuthService', () => {
+  jest.setTimeout(TEST_TIMEOUTS)
+  let containerInstance: ContainerInstance
+
+  beforeAll(async () => {
+    containerInstance = await startContainer()
+  })
+
+  afterAll(async () => await stopContainer(containerInstance))
```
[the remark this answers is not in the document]
beniaminmunteanu: will look into it

packages/backend/src/config/env.ts
beniaminmunteanu: well, i guess the test values will actually be the default ones, since we expect to have them in the docker environment when we want to demo

PR #180, opened 2023-03-10
feat(wallet): KYC Verify Identity image uploads to base64

### LINKED ISSUE #178: Convert to base64 uploaded images for KYc verify
Current Dev implementation sends default base64 image for testing purposes.
WHen not in Dev mode we need to send base64 image and type to /verify.

Tymmmy:
### Context
- fixes #178 

### Changes
Convert uploaded images to base64 and send them to verify identity

### DIFF
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -23,8 +23,8 @@ app.use(
     credentials: true
   })
 )
-app.use(express.json())
-app.use(express.urlencoded({ extended: true }))
+app.use(express.json({ limit: '25mb' }))
+app.use(express.urlencoded({ extended: true, limit: '25mb' }))
 app.use(cookieParser())
 
 const knex = Knex(config[env.NODE_ENV || 'development'])
--- a/packages/wallet/src/components/kyc/VerifyIdentityForm.tsx
+++ b/packages/wallet/src/components/kyc/VerifyIdentityForm.tsx
@@ -8,6 +8,7 @@ import { Form } from '@/ui/forms/Form'
 import { USE_TEST_DATA_KYC } from '@/utils/constants'
 import { getObjectKeys } from '@/utils/helpers'
 import { testImageType, testImageVerifyIdentity } from '@/utils/mocks'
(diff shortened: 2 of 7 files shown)


packages/backend/src/index.ts
```diff
@@ -23,8 +23,8 @@ app.use(
     credentials: true
   })
 )
-app.use(express.json())
-app.use(express.urlencoded({ extended: true }))
```
Tymmmy, line 27: when backside of ID is also being sent to rapyd, we get PayloadTooLargeError, so this is to fix that error

packages/wallet/src/components/kyc/VerifyIdentityForm.tsx
```diff
@@ -16,27 +17,42 @@ import { useKYCFormContext } from './context'
...
-    } else if (target.name === 'frontSideImage') {
-      setFrontIDFile(fileName)
-    } else {
-      setBackIDFile(fileName)
+    if (target.files && target.files.length > 0) {
+      const file = target.files[0]
+      reader.readAsDataURL(file)
+      reader.onloadend = () => {
+        let fileBase64 = reader.result?.toString() || ''
+        fileBase64 = fileBase64.slice(fileBase64.indexOf('base64') + 7)
```
raducristianpopa, line 31: From Rapyd docs:
> ERROR_ID_VERIFICATION_BACK_SIDE_IMAGE_MIME_TYPE_INVALID
The request tried to validate identity, but the back side image provided was not the correct MIME type. The request was rejected. Corrective action: Use a JPEG or PNG image.


It looks like only JPEG or PNG images are allowed (from what I'm seeing this applies to all images, not only `back_side_image` - [reference](https://docs.rapyd.net/build-with-rapyd/reference/identity-verification#identity-verification-errors)). We can add to `FileUpload` component `accepts="image/jpeg, image/png"`, but we will still have to do some extra validation (client and server side). We can create a separate issue for this and address it later on.

raducristianpopa (approved):
🚀

PR #184, opened 2023-03-13
feat(backend): some improvement on create wallet

domin191013:
### Context
some update on create wallet api

### Changes


raducristianpopa:
What do you think if we combine these two queries? Basically update all user columns after the wallet gets created on Rapyd.
https://github.com/interledger/testnet/blob/7a2fa1f28f7c5db38ce6d929f94b2fd7161d988e/packages/backend/src/wallet/wallet.service.ts#L22-L30
https://github.com/interledger/testnet/blob/7a2fa1f28f7c5db38ce6d929f94b2fd7161d988e/packages/backend/src/wallet/wallet.service.ts#L69-L73

dragosp1011 (approved):

PR #190, opened 2023-03-13
fix(wallet): Verify image type for KYC upload

### LINKED ISSUE #185: Image validation for KYC
Rapyd only accepts JPEG and PNG images for ID proof. We should have some extra validation to prevent unnecessary calls to Rapyd API.

References: 
- [Rapyd API Docs](https://docs.rapyd.net/build-with-rapyd/reference/identity-verification#identity-verification-errors)
- https://github.com/interledger/testnet/pull/180#discussion_r1132332696

Tymmmy:
### Context
- fixes #185 

### Changes

### DIFF
--- a/packages/wallet/src/lib/api/user.ts
+++ b/packages/wallet/src/lib/api/user.ts
@@ -38,19 +38,53 @@ export const personalDetailsSchema = z.object({
   zip: z.string().min(1, { message: 'ZIP code is required' })
 })
 
-export const verifyIdentitySchema = z.object({
-  documentType: z.string({
-    invalid_type_error: 'Please select an ID Type'
-  }),
-  frontSideImage: z
-    .string()
-    .min(1, { message: 'Front side of ID is required' }),
-  frontSideImageType: z.string(),
-  backSideImage: z.string().optional(),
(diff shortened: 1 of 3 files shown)


packages/wallet/src/lib/api/user.ts
```diff
@@ -38,19 +38,53 @@ export const personalDetailsSchema = z.object({
...
+    if (!ACCEPTED_IMAGE_TYPES.includes(frontSideImageType.toString())) {
+      ctx.addIssue({
+        code: 'custom',
+        message: `Image must be 'jpeg' or 'png'`,
+        path: ['frontSideImage']
+      })
+    }
+  })
+  .superRefine(({ faceImageType }, ctx) => {
+    if (!ACCEPTED_IMAGE_TYPES.includes(faceImageType.toString())) {
```
raducristianpopa, line 66:
```suggestion
    if (!ACCEPTED_IMAGE_TYPES.includes(faceImageType)) {
```

raducristianpopa (commented):
I think we should be able to do all the validations inside a single `superRefine` instead of calling `superRefine` 3 times.

raducristianpopa (approved):

Issue #194, opened 2023-03-14
Refactor into service and controllers for testing purposes


beniaminmunteanu:
Testing the application would benefit of having the code split up in services and controllers.

This way, tests would no longer need to mock req, res, next when calling the actual methods that need to be tested.
They would call the methods with proper, related params.

example:
<img width="815" alt="image" src="https://user-images.githubusercontent.com/111863110/224933282-9d3139f2-37d4-44cc-9d80-c29519d8c09f.png">

Tymmmy:
New backend scaffold is merged into main, new stories will be opened for separate parts of backend

PR #201, opened 2023-03-17
feat: scaffold new backend

raducristianpopa:
### Context
- context of #194

### Changes

Updates for the `backend` package:

**Container**
- we are now using a container to register and resolve bindings;
- bindings can be registered as singletons or transient (it will return a new instance every time we are trying to resolve a registered binding);

**Structure**
- controllers and services are now separated
- we are using classes to define services and controllers

**Auth**
- switched from `JWT` to [`iron-session`](https://github.com/vvo/iron-session)
- `iron-session` advantages:
  - is stateless (like `JWT`) - we can save user information in the cookie without querying the database to retrieve information about the user.
  - no need to refresh the tokens
  - the data stored in the cookie is encrypted
  - password rotation
  
**Tests**
- we are still using `testcontainers` along with `jest`
- we use `node-mocks-http` to mock `request` and `response`
- tests are isolated - every test suite will now have it's own database (they will not overlap each other)


packages/backend-update/src/auth/router.ts
```diff
@@ -0,0 +1,16 @@
...
+//   constructor(private authController: AuthController) {
+//     this.router = Router()
+//     this.initializeRoutes()
+//   }
+
+//   private initializeRoutes(): void {
+//     this.router.post('/sign-up', this.authController.register)
+//     this.router.post('/log-in', this.authController.login)
+//   }
+// }
```
raducristianpopa, line 16: I don't have a strong opinion against having the routes defined in their own class - my only concern is about having too much abstraction.

What are your opinions? Currently the routes are defined in the `init` method inside `App` class:
https://github.com/interledger/testnet/blob/7a8ffb5542f8fade72bde110360a5849dafd1050/packages/backend-update/src/app.ts#L64-L94

packages/backend-update/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "lib": ["ESNext"],
+    "outDir": "./dist",
+    "rootDir": "./src",
+    "sourceMap": true,
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["./src/*"]
```
raducristianpopa, line 10: I added a custom path to get rid of long import statements (for ie: `../../../../foo/bar`).

packages/backend-update/src/config/env.ts
```diff
@@ -0,0 +1,32 @@
...
+  DATABASE_URL: z
+    .string()
+    .default(
+      'postgres://postgres:password@localhost:5433/testnet-backend-update'
+    ),
+  COOKIE_NAME: z.string().default('testnet.cookie'),
+  COOKIE_PASSWORD: z
+    .string()
+    .default('testnet.cookie.password.super.secret.ilp'), // min. 32 chars
+  COOKIE_TTL: z.coerce.number().default(2630000), // 1 month
```
raducristianpopa, line 15: At this moment, we set the cookie TTL to 1 month. Do we want this to be session based and have the user log in every time it opens the browser?

Tymmmy: I will ask the client on this.

packages/backend-update/src/tests/user/controller.test.ts
```diff
@@ -0,0 +1,116 @@
...
+import { withSession } from '@/middleware/withSession'
+import type { UserService } from '@/user/service'
+import type { UserController } from '@/user/controller'
+
+describe('User Controller', (): void => {
+  let bindings: Container<Bindings>
+  let appContainer: TestApp
+  let knex: Knex
+  let authService: AuthService
+  // let authController: AuthController
```
Tymmmy, line 29: do we need the commented out lines in this file?

packages/backend-update/src/config/env.ts
Tymmmy: I got confirmation that 1 month TTL is ok for testnet. No need to change.

packages/backend-update/src/tests/app.test.ts
```diff
@@ -0,0 +1,48 @@
+import axios, { AxiosError } from 'axios'
```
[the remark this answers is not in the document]
dragosp1011: I would like to have them separately
```
src/
test/
```

packages/backend-update/src/auth/router.ts
dragosp1011: Have a separate file to define all the routes can be a solution

packages/backend-update/src/tests/user/controller.test.ts
raducristianpopa: Nope. Will remove the dead code. 👍

packages/backend-update/src/tests/app.test.ts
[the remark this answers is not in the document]
raducristianpopa: Having them outside the `src` folder, means that we will have to include the folder in `tsconfig.json`.

packages/backend-update/src/auth/router.ts
raducristianpopa: Having them in a separate file I don't think it's a good solution, since everything lives in a container. 🤔

Tymmmy (approved):

packages/backend-update/src/auth/router.ts
beniaminmunteanu: My vote is to keep them in the init method for now. We will figure it out later.

packages/backend-update/src/tests/app.test.ts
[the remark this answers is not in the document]
beniaminmunteanu: I personally prefer having them next to the code. But it's not a strong opinion at all

packages/wallet-backend-update/tests/user/controller.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      expect(res._getJSONData()).toMatchObject({
+        success: true,
+        message: 'User retrieved successfully',
+        data: {
+          email: args.email
+        }
+      })
+    })
+
+    it('it returns 404 if the session is not found', async (): Promise<void> => {
```
dragosp1011, line 93: duplication for it - maybe we can change the format to`it('should return 404 if the session is not found'`

packages/wallet-backend-update/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+import { AuthController } from './auth/controller'
+import { AuthService } from './auth/service'
+import { Env, env } from './config/env'
+import { logger } from './config/logger'
+import { Container } from './container'
+import { UserService } from './user/service'
+import { UserController } from './user/controller'
+import { SessionService } from './session/service'
+
+export const createContainer = (config: Env): Container<Bindings> => {
```
dragosp1011, line 12: I suggest to extract this method in a separate file

Tymmmy (approved):

packages/wallet-backend-update/tests/auth/controller.test.ts
```diff
@@ -0,0 +1,192 @@
+import { createContainer } from '@/createContainer'
+import { Bindings } from '@/app'
+import { env } from '@/config/env'
+import { Container } from '@/shared/container'
+import { createApp, TestApp } from '../app'
```
rico191013, line 5: How about add tests to tsconfig as well? So that this can be @/tests/app

packages/wallet-backend-update/tests/auth/controller.test.ts
```diff
@@ -0,0 +1,192 @@
...
+  createRequest,
+  createResponse,
+  MockRequest,
+  MockResponse
+} from 'node-mocks-http'
+import type { AuthController } from '@/auth/controller'
+import type { AuthService } from '@/auth/service'
+import { errorHandler } from '@/middleware/errorHandler'
+import { faker } from '@faker-js/faker'
+import { applyMiddleware } from '../utils'
```
rico191013, line 19: I think it'd be good if we use same format for file path. '@/app/...'

packages/wallet-backend-update/tests/auth/controller.test.ts
```diff
@@ -0,0 +1,192 @@
...
+  let authController: AuthController
+  let userService: UserService
+  let req: MockRequest<Request>
+  let res: MockResponse<Response>
+
+  const next = jest.fn()
+  const args = {
+    email: faker.internet.email(),
+    password: faker.internet.password()
+  }
```
rico191013, line 37: Could we create a mock file, and put args in it? So that it can be imported in several test files..

rico191013 (changes requested):

packages/wallet-backend-update/src/app.ts
```diff
@@ -0,0 +1,116 @@
...
+    const router = Router()
+
+    const env = await this.container.resolve('env')
+    const authController = await this.container.resolve('authController')
+    const userController = await this.container.resolve('userController')
+
+    app.use(helmet())
+    app.use(express.json())
+    app.use(express.urlencoded({ extended: true, limit: '25mb' }))
+    app.use(withSession)
```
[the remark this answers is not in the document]
raducristianpopa: The `withSession` middleware simply attaches the `session` object (`req.session`) to the incoming request, while the `isAuth` middleware is responsible for verifying user authentication.

packages/wallet-backend-update/src/user/controller.ts
```diff
@@ -0,0 +1,65 @@
...
+  lastName?: string
+  firstName?: string
+  address?: string
+  country?: string
+  needsWallet: boolean
+  needsIDProof: boolean
+}
+
+interface IUserController {
+  me: (
```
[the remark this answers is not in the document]
raducristianpopa: The `/me` endpoint can be hit even if the user is not authenticated - it is not guarded by the `isAuth` middleware.
It has the same behaviour like the current `/me` endpoint on our initial backend - retrieving user information if it exists. We call `session.destroy` to delete all the cookies that the user has - in case there are any.

On the other hand, I think we can simply improve this by returning early if `req.session.id` or `req.session.user` do not exist. By doing this we will not make queries to our database if the session is invalid.

packages/wallet-backend-update/tests/auth/controller.test.ts
raducristianpopa: I like this approach and I'll look into it. Let's hope TS will not yell at us.

packages/wallet-backend-update/src/app.ts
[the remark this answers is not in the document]
beniaminmunteanu: In that case, shouldn't the isAuth middleware be applied to the '/me' route here at line 92?

packages/wallet-backend-update/src/user/controller.ts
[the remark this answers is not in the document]
beniaminmunteanu: Wouldn't it be better to have it guarded by IsAuth, in which case it would return Unauthorised if the req.session.id or req.session.user do not exist?
 
 I don't see a usecase in which you hit the me endpoint as an unauthorised user?

dragosp1011 (approved):

rico191013 (approved):

beniaminmunteanu:
:muscle::bear:

dragosp1011 (approved):

Tymmmy (approved):

PR #282, opened 2023-04-21
feat: Use debounced input for IPUrl, and populate amount, description

### LINKED ISSUE #262: Read Incoming payment info and auto populate inputs
On the Pay page we paste the incoming payment url, we should also auto populate amount and all the info from the incoming payment url.

Tymmmy:
### Context
- fixes #262 

### Changes
- added debounce input to Incoming Payment Url
- made GET API call for IPYrl details
- on success populate amound & description automatically


packages/wallet-frontend/src/utils/debounce.ts
```diff
@@ -0,0 +1,16 @@
...
+    const handler = setTimeout(() => {
+      setDebouncedValue(value)
+    }, delay)
+
+    return () => {
+      clearTimeout(handler)
+    }
+  }, [value, delay])
+  return debouncedValue
+}
```
raducristianpopa, line 16: I personally tend to stay away from default exports unless they are really needed
```suggestion
import { useEffect, useState } from 'react'

function useDebounce(value: string, delay: number) {
  const [debouncedValue, setDebouncedValue] = useState(value)

  useEffect(() => {
    const handler = setTimeout(() => {
      setDebouncedValue(value)
    }, delay)

    return () => {
      clearTimeout(handler)
    }
  }, [value, delay])
  return debouncedValue
}

export { useDebounce }
```

packages/wallet-frontend/src/utils/debounce.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react'
+
+export default function useDebounce(value: string, delay: number) {
```
raducristianpopa, line 3: A more generic approach. This way we can use any type for `value`, not only `string`.

```suggestion
export default function useDebounce<T>(value: T, delay: number) {
```

packages/wallet-frontend/src/utils/debounce.ts
```diff
@@ -0,0 +1,16 @@
...
+  useEffect(() => {
+    const handler = setTimeout(() => {
+      setDebouncedValue(value)
+    }, delay)
+
+    return () => {
+      clearTimeout(handler)
+    }
+  }, [value, delay])
+  return debouncedValue
```
raducristianpopa, line 15: What do you think about storing the previous value? In our case, if a user pastes the same URL into the input it would not make another call to the backend.

packages/wallet-frontend/src/utils/debounce.ts
Tymmmy: Good idea

packages/wallet-frontend/src/utils/debounce.ts
Tymmmy: ALso a great idea, will try to implement, but the previous value should be verified in the pay.tsx and not in the debounce, no?

raducristianpopa: It would make sense to be in the `useDebounce` hook. Thinking more about this, we can also return a `isLoading` property to display a loading state while we are waiting for the debounced value.

packages/wallet-frontend/src/lib/api/asset.ts
```diff
@@ -14,10 +14,13 @@ export type Asset = {
 }
 
 type ListAssetsResult = SuccessResponse<Asset[]>
-type ListAssetsResponse = Promise<ListAssetsResult | ErrorResponse>
+type GetAssetResult = SuccessResponse<Asset>
+type ListAssetsResponse = ListAssetsResult | ErrorResponse
+type GetAssetsResponse = GetAssetResult | ErrorResponse
```
raducristianpopa, line 19:
```suggestion
type GetAssetResponse = GetAssetResult | ErrorResponse
```

raducristianpopa, line 19: The types are mixed up. Let's sort them for easier reading.

packages/wallet-frontend/src/lib/api/asset.ts
```diff
@@ -14,10 +14,13 @@ export type Asset = {
...
 
 interface AssetService {
   list: (cookies?: string) => Promise<ListAssetsResponse>
+  get: (id: string, cookies?: string) => Promise<GetAssetsResponse>
```
raducristianpopa, line 23:
```suggestion
  get: (id: string, cookies?: string) => Promise<GetAssetResponse>
```

packages/wallet-frontend/src/lib/api/asset.ts
```diff
@@ -34,6 +37,20 @@ const createAssetService = (): AssetService => ({
     } catch (error) {
       return getError(error, 'Unable to fetch assets.')
     }
+  },
+
+  async get(id, cookies): Promise<GetAssetsResponse> {
```
raducristianpopa, line 42:
```suggestion
  async get(id, cookies): Promise<GetAssetResponse> {
```

packages/wallet-frontend/src/lib/api/transfers.ts
```diff
@@ -45,23 +45,32 @@ type SendArgs = z.infer<typeof sendSchema>
 type SendError = ErrorResponse<SendArgs | undefined>
 type SendResponse = Promise<SuccessResponse | SendError>
 
-type RequestSuccessResponse = {
+type TransactionResponse = {
```
raducristianpopa, line 48: I think we can rename this to `Transaction` only? 🤔
```suggestion
type Transaction = {
```

packages/wallet-frontend/src/lib/hooks/useDebounce.ts
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useRef, useState } from 'react'
+
+function useDebounce<T>(value: T, delay: number) {
+  const [debouncedValue, setDebouncedValue] = useState(value)
+
+  const lastValue = useRef<T>(value)
+  const [isLoading, setIsLoading] = useState(false)
+
+  useEffect(() => {
+    if (value !== '' && lastValue !== value) {
```
raducristianpopa, line 10: I think this should be:
```suggestion
    if (value !== '' && lastValue.current !== value) {
```

packages/wallet-frontend/src/lib/hooks/useDebounce.ts
```diff
@@ -0,0 +1,26 @@
...
+      setIsLoading(true)
+      const handler = setTimeout(() => {
+        setIsLoading(false)
+        setDebouncedValue(value)
```
raducristianpopa, line 14: The proper order would be `setDebouncedValue` and `setIsLoading` after, but this works as well.

packages/wallet-frontend/src/lib/hooks/useDebounce.ts
```diff
@@ -0,0 +1,26 @@
...
+      }, delay)
+
+      return () => {
+        setIsLoading(false)
```
raducristianpopa, line 18: Is this needed? Wouldn't `setIsLoading` that is placed inside `handler` cover everything?

packages/wallet-frontend/src/pages/transfer/pay.tsx
```diff
@@ -13,26 +13,70 @@ import { accountService } from '@/lib/api/account'
...
+import { useEffect, useState } from 'react'
 import { paymentPointerService } from '@/lib/api/paymentPointer'
 import { ErrorDialog } from '@/components/dialogs/ErrorDialog'
+import { useDebounce } from '@/lib/hooks/useDebounce'
+import { assetService } from '@/lib/api/asset'
 
 type PayProps = InferGetServerSidePropsType<typeof getServerSideProps>
 
 export default function Pay({ accounts }: PayProps) {
+  let scale = 2
```
raducristianpopa, line 26: Wouldn't it be better to return the transformed amount directly from backend? This way we would get rid of some logic in the client side.

packages/wallet-frontend/src/pages/transfer/pay.tsx
```diff
@@ -13,26 +13,70 @@ import { accountService } from '@/lib/api/account'
...
+      payForm.setValue('description', '')
+    }
+  }
+
+  useEffect(() => {
+    if (debouncedIncomingPaymentUrl) {
+      fetchData()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [debouncedIncomingPaymentUrl])
```
raducristianpopa, line 63: How about using a `DebouncedInput` component instead? That way, we might be able to eliminate this `useEffect`. 🤔

packages/wallet-frontend/src/pages/transfer/pay.tsx
```diff
@@ -14,25 +14,52 @@ import { paySchema, transfersService } from '@/lib/api/transfers'
...
+    const incomingPaymentId = tempArray[tempArray.length - 1]
+    const response = await transfersService.getIncomingPaymentDetails(
+      incomingPaymentId
+    )
+    if (response.success && response.data) {
+      const { value, description } = response.data
+      payForm.clearErrors('root')
+      payForm.setValue('amount', value)
+      payForm.setValue('description', description ?? '')
+      payForm.setValue('incomingPaymentUrl', incomingPaymentUrl)
```
raducristianpopa, line 50: When we assign a value to the `incomingPaymentUrl` field, is it trimmed automatically? I noticed that the `zod` schema uses the `trim` function, but I'm uncertain if this is reflected visually.

packages/wallet-frontend/src/components/PaymentPointerCard.tsx
```diff
@@ -128,8 +128,7 @@ export const PaymentPointerCard = ({
           <Form
             className="w-full"
             form={form}
-            onSubmit={(data) => {
-              console.log(data)
+            onSubmit={() => {
```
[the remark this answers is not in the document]
Tymmmy: I think it is ok to just remove it here, since we discovered it.

raducristianpopa (approved):

PR #292, opened 2023-04-24
feat(wallet): add soft delete functionality

### LINKED ISSUE #247: Remove Edit, Hide from view (soft delete) Payment Pointers Functionality
Remove edit functionality and finish soft delete from view (hide from PP list) functionality for payment pointers.

image

raducristianpopa:
### Context
- fixes #247 

### Changes
- Edit functionality is no longer available
- Added `active` column to `paymentPointer` table
- Deletion of a payment pointer is now working as a soft delete

When the user wants to delete a payment pointer, we update the `active` column to `false` (soft delete - user will not see the payment pointer in the UI). If the user creates a new payment pointer with the same name, we will update the `active` to `true`.


Tymmmy (dismissed):

packages/wallet-backend/migrations/20230209095526_payment_pointer.js
```diff
@@ -9,6 +9,7 @@ exports.up = function (knex) {
     table.string('publicName').notNullable()
     table.uuid('accountId').notNullable()
     table.foreign('accountId').references('accounts.id')
+    table.boolean('active').notNullable().defaultTo(true)
```
dragosp1011, line 12: maybe `isActive`

packages/wallet-backend/src/payment-pointer/payment-pointer.service.ts
```diff
@@ -23,29 +23,41 @@ export const createPaymentPointer = async (
...
+    if (paymentPointer) {
+      if (paymentPointer.accountId != accountId || account.userId !== userId) {
+        throw new ConflictException(
+          'This payment pointer already exists. Please choose another name.'
+        )
+      } else if (
+        paymentPointer.accountId === accountId &&
+        account.userId === userId
+      ) {
+        paymentPointer = await PaymentPointerModel.query().patchAndFetchById(
```
dragosp1011, line 39: it can be confusing if you have transactions on the payment pointer (you just created a payment pointer but you have transactions on it)

dragosp1011 (dismissed):

packages/wallet-backend/src/payment-pointer/payment-pointer.service.ts
raducristianpopa: You are right, it can be confusing to create a new payment pointer and already have transactions for it. Let's merge this in and create a separate issue for this and discuss it further.

Tymmmy (approved):

PR #322, opened 2023-04-28
feat(transactions): add default sort order by created at

### LINKED ISSUE #309: Update transactions table
We should display the transactions in descending order - newest to oldest, by default.

Another good addition might be to show the date and time for each transaction, not only the date.
@Tymmmy thoughts on this? 👆

dragosp1011:
### Context
- fixes #309

### Changes

### DIFF
--- a/packages/wallet-backend/src/middlewares/validator.ts
+++ b/packages/wallet-backend/src/middlewares/validator.ts
@@ -4,9 +4,10 @@ import { BadRequestException } from '../shared/models/errors/BadRequestException
 
 export async function zParse<T extends AnyZodObject>(
   schema: T,
-  req: Request
+  req: Request,
+  property: 'body' | 'query' = 'body'
 ): Promise<z.infer<T>> {
-  const res = await schema.safeParseAsync(req.body)
+  const res = await schema.safeParseAsync(req[property])
   if (!res.success) {
     const errors: Record<string, string> = {}
     res.error.issues.forEach((i) => {
--- a/packages/wallet-frontend/src/pages/account/[accountId]/[paymentPointerId].tsx
+++ b/packages/wallet-frontend/src/pages/account/[accountId]/[paymentPointerId].tsx
@@ -137,7 +137,6 @@ export const getServerSideProps: GetServerSideProps<{
     }
   }
 
-  console.log(transactionsResponse.data)
   const transactions = transactionsResponse.data.map((trx) => ({
     ...trx,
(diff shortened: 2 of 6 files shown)


packages/wallet-backend/src/middlewares/validator.ts
```diff
@@ -4,9 +4,10 @@ import { BadRequestException } from '../shared/models/errors/BadRequestException
 
 export async function zParse<T extends AnyZodObject>(
   schema: T,
-  req: Request
+  req: Request,
+  property: 'body' | 'query' = 'body'
```
raducristianpopa, line 8:
```suggestion
  req: Request
```

packages/wallet-backend/src/middlewares/validator.ts
```diff
@@ -4,9 +4,10 @@ import { BadRequestException } from '../shared/models/errors/BadRequestException
...
 ): Promise<z.infer<T>> {
-  const res = await schema.safeParseAsync(req.body)
+  const res = await schema.safeParseAsync(req[property])
```
raducristianpopa, line 10:
```suggestion
  const res = await schema.safeParseAsync(req)
```

raducristianpopa (commented):
LGTM. Looking more into it, I think we can improve the request validator.
Here is what I am thinking:
- pass the whole request object to the validator
- create validations schema for `body`, `query` and `params` in the same schema

The validation schema would look like this:
```ts
export const transactionListRequestSchema = z.object({
  body: z.object({ ... }), // if we expect a body
  query: z.object({ ... }), // if we expect query string parameters
  params: z.object({ ... }), // if we expect route parameters
})
```

What do you think about this?

Tymmmy (approved):

Tymmmy (approved):

packages/wallet-frontend/src/pages/account/[accountId]/[paymentPointerId].tsx
```diff
@@ -142,7 +142,9 @@ export const getServerSideProps: GetServerSideProps<{
     createdAt: new Date(trx.createdAt).toLocaleDateString('default', {
       day: '2-digit',
       month: 'short',
-      year: 'numeric'
+      year: 'numeric',
+      hour: '2-digit',
+      minute: '2-digit'
```
Comment (raducristianpopa) [inline packages/wallet-frontend/src/pages/account/[accountId]/[paymentPointerId].tsx:147]: Should we relocate this transformation in the backend?

PR #337, opened 2023-05-03
feat(wallet/frontend): update input component 

### LINKED ISSUE #323: Change input component
Replace the current input component with one that is matching this layout
Input

raducristianpopa:
### Context
- fixes #323

### Changes
- Farewell floating labels; 🥲
- A new input component has been introduced which supports add-on and trailing content;
- Labels have been moved outside the input component;
- Labels can now include hints.

<details>
  <summary>Features preview:</summary>
  
image

</details>


packages/wallet-frontend/src/ui/forms/Input.tsx
```diff
@@ -1,29 +1,56 @@
...
 import { FieldError } from './FieldError'
+import { cx } from 'class-variance-authority'
 
-type InputProps = Omit<ComponentPropsWithoutRef<'input'>, 'className'> & {
+type InputProps = ComponentPropsWithoutRef<'input'> & {
   label?: string
   error?: string
+  addOn?: ReactNode
+  trailing?: ReactNode
+  labelHint?: ReactNode
```
raducristianpopa, line 12: I've been thinking of adding a `trailingIcon` or `isLoading` property. My though process:
- there is some async work that needs to be done for a certain input - the `isLoading` case
- we might want to add a trailing icon (error badge, show/hide for passwords) - the `trailingIcon` case

My main concern would be if we have trailing content and a trailing icon. Should we only allow  one if the trailing properties (ie: if `trailing` is present, do not accept `trailingIcon` and the other way around)?

I have been considering the possibility of implementing a `trailingIcon` or `isLoading` property. My reasoning behind this is as follows:
- there may be certain inputs that require asynchronous work - the `isLoading` case
- it may be useful to include a trailing icon (such as an error badge or a show/hide button for password) - the `trailingIcon` case

My primary concern is whether or not we should allow both trailing content and a trailing icon. Should we only permit one of these trailing options (e.g., if `trailing` is present, we do not accept `trailingIcon` and vice versa)?

raducristianpopa, line 12: I have been considering the possibility of implementing a `trailingIcon` or `isLoading` property. My reasoning behind this is as follows:
- there may be certain inputs that require asynchronous work - the `isLoading` case
- it may be useful to include a trailing icon (such as an error badge or a show/hide button for password) - the `trailingIcon` case

My primary concern is whether or not we should allow both trailing content and a trailing icon. Should we only permit one of these trailing options (e.g., if `trailing` is present, we do not accept `trailingIcon` and vice versa)?

Tymmmy: It would be very unusual to have both at the same time indeed, but not impossible. I would think it is ok to allow both at the same time...

packages/wallet-frontend/src/components/kyc/PersonalDetailsForm.tsx
```diff
@@ -100,6 +100,8 @@ export const PersonalDetailsForm = () => {
         name="country"
         render={({ field: { value } }) => (
           <Select<SelectOption>
+            required
+            label="Country"
             options={countries}
```
Tymmmy, line 105: Should have maybe a placeholder of Select country... to be consistent with other selects

packages/wallet-frontend/src/pages/account/create.tsx
```diff
@@ -56,7 +56,6 @@ export default function CreateAccount({ assets }: CreateAccountProps) {
       >
         <Input
           required
-          placeholder="My Account"
           label="Account name"
           error={createAccountForm.formState?.errors?.name?.message}
           {...createAccountForm.register('name')}
```
Tymmmy, line 61: Should add to the Select in this file a label (Asset or Currency) and a placeholder (Select asset or currency...)

packages/wallet-frontend/src/ui/forms/Select.tsx
```diff
@@ -84,13 +95,14 @@ export const Select = <
         classNames={{
           control: ({ isDisabled, isFocused }) =>
             cx(
-              'rounded-xl shadow-md border',
-              isDisabled ? 'bg-gray-200' : 'bg-white',
+              'rounded-md shadow-md border py-0.',
```
Tymmmy, line 98: is that '.' intended to be there? a 0. ?

Tymmmy (approved):

PR #353, opened 2023-05-10
feat(wallet-backend): Some code changes for GCP deployment 

cryptofan1013:
### Context
- fixes #

### Changes


docker/dev/docker-compose-gcp.yml
```diff
@@ -0,0 +1,194 @@
...
+      context: ../..
+      dockerfile: ./packages/wallet-backend/Dockerfile.dev
+    depends_on:
+      - postgres
+      - rafiki-backend
+    environment:
+      NODE_ENV: development
+      PORT: 3003
+      DB_URL: postgres://wallet_backend:wallet_backend@postgres/wallet_backend
+      JWT_ACCESS_TOKEN_SECRET: u3uakiThYILFEewwhlHkepwMLM@Ikbj1LA!JQeeVyVEn$$Cj9t$$Ix0TWMYrEnpq$$X
```
dragosp1011, line 49: I suggest to add the jwt tokens as it is done for rapyd

packages/wallet-frontend/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "wallet-frontend",
   "scripts": {
     "dev": "PORT=4003 next dev",
-    "build": "next build",
+    "build": "NEXT_PUBLIC_BACKEND_URL=http://35.196.11.156:3003 next build",
```
dragosp1011, line 5: is this change required?

cryptofan1013: well, the env values in docker-compose.yml wasn't affecting the build environment variables.
Which means, without this, that backend url was setting to another value.

docker/dev/docker-compose-gcp.yml
beniaminmunteanu: I don't understand :)

packages/wallet-frontend/package.json
beniaminmunteanu: I also find it a bit weird to add the env variables here in the build script

docker/dev/docker-compose-gcp.yml
dragosp1011: add them in .env file and here change it to `JWT_ACCESS_TOKEN_SECRET: ${JWT_ACCESS_TOKEN_SECRET}`

packages/wallet-backend/src/index.ts
```diff
@@ -23,7 +23,15 @@ app.disable('X-Powered-By')
 
 app.use(
   cors({
-    origin: 'http://localhost:4003',
+    origin: [
+      'http://localhost:4003',
```
dragosp1011, line 27: should these be in env file?

packages/wallet-frontend/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "wallet-frontend",
   "scripts": {
     "dev": "PORT=4003 next dev",
-    "build": "next build",
+    "build": "NEXT_PUBLIC_BACKEND_URL=https://rafiki.money/api next build",
```
dragosp1011, line 5: the env can be set in docker on running the build script

docker/dev/docker-compose-gcp.yml
cryptofan1013: I think this can be solved after it's updated in main docker file? not docker file for gcp.

packages/wallet-backend/src/index.ts
cryptofan1013: Well, I think as the domain will be fixed, it can be here as well.

docker/dev/docker-compose-prod.yml
```diff
@@ -0,0 +1,194 @@
...
+    container_name: wallet-frontend
+    build:
+      context: ../..
+      dockerfile: ./packages/wallet-frontend/Dockerfile-Prod.dev
+    environment:
+      NODE_ENV: development
+      PORT: 4003
+      NEXT_PUBLIC_USE_TEST_KYC_DATA: 'true'
+      NEXT_PUBLIC_BACKEND_URL: https://api.rafiki.money
+      NEXT_PUBLIC_OPEN_PAYMENTS_HOST: http://rafiki-backend
```
dragosp1011, line 29: is this value correct?

Tymmmy:
Looks like you need to run prettier also  on auth.service file

docker/dev/docker-compose-prod.yml
```diff
@@ -0,0 +1,194 @@
...
+    build:
+      context: ../..
+      dockerfile: ./packages/wallet-backend/Dockerfile.dev
+    depends_on:
+      - postgres
+      - rafiki-backend
+    environment:
+      NODE_ENV: development
+      PORT: 3003
+      DB_URL: postgres://wallet_backend:wallet_backend@postgres/wallet_backend
```
dragosp1011, line 48: DB urls and jwt secrets should also be in .env IMO

beniaminmunteanu (approved):

packages/wallet-backend/src/auth/auth.service.ts
```diff
@@ -39,7 +39,10 @@ export const appendAccessTokenToCookie = (
   expiresIn: number
 ) => {
   res.cookie('AccessToken', token, {
-    httpOnly: true,
+    httpOnly: env.NODE_ENV === 'production',
```
dragosp1011, line 42: extract cookies options in one common object/method

packages/wallet-backend/src/auth/auth.service.ts
```diff
@@ -39,7 +39,10 @@ export const appendAccessTokenToCookie = (
...
+    secure: env.NODE_ENV === 'production',
+    sameSite: 'none',
+    domain: env.NODE_ENV === 'production' ? 'rafiki.money' : 'localhost:4003',
```
dragosp1011, line 45: I think domain should be in an env variable, that will allow to setup other environments without changing the code

dragosp1011 (changes requested):

packages/wallet-backend/src/auth/auth.service.ts
cryptofan1013: well, this is for wallet-backend only. In wallet-backend-update, there's only one option for cookies already.

dragosp1011 (approved):

Issue #329, opened 2023-04-28
Validate request body, query and params in one schema


raducristianpopa:
Only for the new backend.
Context: https://github.com/interledger/testnet/pull/322#pullrequestreview-1405708096

PR #355, opened 2023-05-11
feat: Integrate features in new backend

beniaminmunteanu:
**This PR moves the old backend features to the new backend structure and DI**


packages/wallet-backend-update/src/account/schemas/account.schema.ts
raducristianpopa, line 1: This should live in a `validation.ts` file in `./src/account`.

packages/wallet-backend-update/src/account/schemas/fund.schema.ts
raducristianpopa, line 1: Ditto above.

packages/wallet-backend-update/src/app.ts
```diff
@@ -12,23 +12,49 @@ import type { AuthController } from './auth/controller'
...
 export interface Bindings {
   env: Env
   logger: Logger
   knex: Knex
+  gqlClient: GraphQLClient
+  rapydClient: RapydClient
+  rafikiClient: RafikiClient
+  rapydCountriesService: RapydCountriesService
+  rapydDocumentsService: RapyddocumentsService
+  rapydWalletService: RapydWalletService
```
raducristianpopa, line 44: Everything related to Rapyd should live in `rapydService` and everything related to Rafiki should live in `rafiki`/`rafikiService`. I don't think we need to expose the GraphQL Client and can only live in the rafiki service.

packages/wallet-backend-update/src/asset/schemas/asset.schema.ts
raducristianpopa, line 1: Ditto - `validation.ts` file.

packages/wallet-backend-update/src/index.ts
```diff
@@ -1,8 +1,12 @@
+import { GraphQLClient } from 'graphql-request'
 import { App } from './app'
 import { env } from './config/env'
 import { createContainer } from './createContainer'
 
 const container = createContainer(env)
+//! For now, until we determine if this needs to be dependency injected as well.
+export const graphqlClient = new GraphQLClient(env.GRAPHQL_ENDPOINT)
+
```
raducristianpopa, line 9: See my comment above.

packages/wallet-backend-update/src/rafiki/request/asset.request.ts
raducristianpopa, line 1: Instead of having separate files, we can have `queries.ts` and `mutations.ts` where we can store the GraphQL documents.

packages/wallet-backend-update/src/rapyd/countries/service.ts
raducristianpopa, line 1: See my comment above regarding Rapyd.

packages/wallet-backend-update/src/rapyd/rapyd.d.ts
raducristianpopa, line 1: It's unnecessary to have every precise property for each type. We could either proceed with this approach (having types only for the required properties) or construct a `zod` validation schema using only the required properties.

packages/wallet-backend-update/src/rapyd/wallet/controller.ts
raducristianpopa, line 1: I think these should be moved into `accountService`.

packages/wallet-backend-update/src/rapyd/wallet/service.ts
```diff
@@ -0,0 +1,177 @@
...
+    firstName: string,
+    lastName: string,
+    address: string,
+    city: string,
+    country: string,
+    zip: string,
+    email: string,
+    id: string,
+    phone?: string
+  ) {
```
raducristianpopa, line 29: We should have an object for this instead of 9 arguments.

packages/wallet-backend-update/src/rapyd/wallet/service.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  public async verifyIdentity(
+    userId: string,
+    documentType: string,
+    frontSideImage: string,
+    frontSideImageType: string,
+    faceImage: string,
+    faceImageType: string,
+    backSideImage?: string,
+    backSideImageType?: string
```
raducristianpopa, line 98: Ditto above.

raducristianpopa (commented):
Considering this is the initial pull request integrating the features into the updated codebase, it should also provide us with the opportunity to address issue #329 within the same context.

packages/wallet-backend-update/src/account/controller.ts
```diff
@@ -0,0 +1,125 @@
...
+import type { NextFunction, Request } from 'express'
+import type { Logger } from 'winston'
+import { validate } from '../shared/validate'
+import { accountSchema } from './schemas/account.schema'
+import { AccountService } from './service'
+import { fundSchema } from './schemas/fund.schema'
+import { Account } from './model'
+
+interface IAccountController {
+  createAccount: (
```
dragosp1011, line 11: proposal: we can have a general controller function type
```
type ControllerFunction<T = undefined> = (
    req: Request,
    res: CustomResponse<T>,
    next: NextFunction
) => Promise<void>
```

packages/wallet-backend-update/src/app.ts
beniaminmunteanu: I would not tie together the rafiki and graphql client, as future proof in case other graphql api might be used, we could use the same gql service.

Everything that makes part of rapyd should be in the rapyd service, indeed.

packages/wallet-backend-update/src/rafiki/request/asset.request.ts
beniaminmunteanu: how does the codegen look for these? is it based on the .request.ts filename? Or something else

packages/wallet-backend-update/src/app.ts
beniaminmunteanu: Also I want to add that I didn't move all the rapyd related stuff together is so that I keep the previous routes that Rico had, so there would be no changes to the frontend yet. 
If we move them together, I think this PR also needs to adress the frontend url's used in requests

packages/wallet-backend-update/src/rafiki/request/asset.request.ts
raducristianpopa: https://github.com/interledger/testnet/blob/b748fad4acf07ae111129fac0e0547d9c3fe8a42/packages/wallet-backend/codegen.yml#L3

packages/wallet-backend-update/src/incomingPayment/service.ts
```diff
@@ -0,0 +1,118 @@
...
+import { NotFound } from '@/errors'
+import { extractUuidFromUrl, transformAmount } from '@/utils/helpers'
+import { Asset } from '@/rafiki/generated/graphql'
+
+interface IIncomingPaymentService {
+  create: (
+    userId: string,
+    paymentPointerId: string,
+    amount: number,
+    description: string
```
raducristianpopa, line 15: Ditto above.

packages/wallet-backend-update/src/incomingPayment/validation.ts
```diff
@@ -0,0 +1,25 @@
...
+})
+
+export const paymentDetailsSchema = z.object({
+  query: z.object({
+    url: z
+      .string()
+      .url()
+      .regex(
+        new RegExp(
+          /\/[a-z1-9_]*\/incoming-payments\/[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/
```
raducristianpopa, line 18: I think we should decode the query parameter first.

packages/wallet-backend-update/src/rafiki/controller.ts
```diff
@@ -0,0 +1,55 @@
...
+interface RafikiControllerDependencies {
+  logger: Logger
+  rafikiService: RafikiService
+}
+
+export class RafikiController implements IRafikiController {
+  constructor(private deps: RafikiControllerDependencies) {}
+  createQuote = async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const receivedQuote = req.body
```
raducristianpopa, line 26: Shouldn't we validate the `body` before passing it to the service?

packages/wallet-backend-update/src/rafiki/controller.ts
```diff
@@ -0,0 +1,55 @@
...
+    try {
+      res.status(200).json(this.deps.rafikiService.getRates())
+    } catch (e) {
+      next(e)
+    }
+  }
+
+  onWebHook = async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const wh = req.body
```
raducristianpopa, line 48: Ditto above.

packages/wallet-backend-update/src/rapyd/rapyd-client.ts
```diff
@@ -0,0 +1,210 @@
...
+        // await handleSettlementOutOfFunds(req, env.RAPYD_SETTLEMENT_EWALLET)
+      }
+      throw new Error(transferResponse.status.message)
+    }
+
+    const setTransferResponse = await this.setTransferResponse({
+      id: transferResponse.data.id,
+      status: 'accept'
+    })
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
raducristianpopa, line 88: We should type the responses.

packages/wallet-backend-update/src/rapyd/rapyd-client.ts
raducristianpopa, line 1: Types for the Rapyd client are missing in the most part. A good approach would be to hold the producer accountable for the types instead of the consumer.

packages/wallet-backend-update/src/rapyd/rapyd.d.ts
raducristianpopa, line 1: We should decide if we want to stick with declaration files or not. Having only the required types instead of listing them all would be a good approach as well.

packages/wallet-backend-update/src/rapyd/service.ts
```diff
@@ -0,0 +1,206 @@
...
+          phone_number: params.phone
+        }
+      }
+    })
+
+    if (result.status.status !== 'SUCCESS') {
+      //!THROW
+      // this.deps.logger
+      // throw new InternalServerError()
+    }
```
raducristianpopa, line 112: What should happen in case of an error?

packages/wallet-backend-update/src/utils/helpers.ts
```diff
@@ -0,0 +1,20 @@
+export const formatBalance = (value: bigint | number, scale = 2): bigint => {
+  return BigInt(transformAmount(value, scale).toFixed(2))
+}
+
+export const transformAmount = (value: bigint | number, scale = 2) => {
+  return Number(BigInt(value) / BigInt(10 ** scale))
+}
```
raducristianpopa, line 7: I don't think `scale` should have a default value

packages/wallet-backend-update/src/account/service.ts
```diff
@@ -0,0 +1,200 @@
...
+        )?.balance ?? 0
+      )
+    })
+
+    return accounts
+  }
+
+  public async getAccountById(
+    userId: string,
+    accountId: string
```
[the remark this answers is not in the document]
beniaminmunteanu: I wouldn't necessarily go for params sent in objects for all methods, especially when only 2-3 params are sent. Linter does a good job of always infering types, no need for extra defined Types of objects that need to be sent as params.

However, I do agree that it would be nice to have the same style everywhere, so I will wait for @dragosp1011  and @rico191013 's opinion on this so that we decide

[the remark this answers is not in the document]
dragosp1011: agree with Beni, for more than 2-3 params we can use object. If all the params are like a logical structure (a payment payload for example) then it can be an object even if there are 2-3 params.

packages/wallet-backend-update/src/incomingPayment/validation.ts
beniaminmunteanu: ?

packages/wallet-backend-update/src/rapyd/rapyd-client.ts
beniaminmunteanu: could you elaborate?

beniaminmunteanu: I was thinking of going through the Rapyd docs and typing the responses, since that is what we are mostly lacking right now. We have the rapyd requests typed out for the rapyd client, but not the responses.

packages/wallet-backend-update/src/rapyd/rapyd.d.ts
beniaminmunteanu: you mean only having the required properties in rapyd request types?

beniaminmunteanu: what about rapyd responses, see my comment above ^^

packages/wallet-backend-update/src/rapyd/service.ts
beniaminmunteanu: create wallet should fail entirely.

packages/wallet-backend-update/src/utils/helpers.ts
beniaminmunteanu: @dragosp1011

packages/wallet-backend-update/src/incomingPayment/validation.ts
raducristianpopa: Since the payment URL will come as a query parameter, some characters will be encoded. If `express` does not automatically decode them, we should decode them (e.g.: in our use case it will be a payment URL, so the query parameter will be encoded to `http%3A%2F%2Flocalhost%2Fincoming...`).

packages/wallet-backend-update/src/rapyd/rapyd-client.ts
raducristianpopa: > could you elaborate?

Instead of typing the value when using it, we should type the function return type.
E.g.:

```ts
public async createWallet(params: CreateWalletParams): ReturnType { ... }
```

https://github.com/interledger/testnet/blob/57833628171c5ba2c5bce905aa78e28d88d380cc/packages/wallet-backend-update/src/rapyd/controller.ts#L68-L78
Here the type of `createWalletResponse` is any.

packages/wallet-backend-update/src/incomingPayment/validation.ts
beniaminmunteanu: @dragosp1011  can you check this?

dragosp1011: @raducristianpopa I think it is decoded as it is working now

packages/wallet-backend-update/src/rafiki/controller.ts
```diff
@@ -0,0 +1,58 @@
...
+  createQuote: (
+    req: Request,
+    res: Response<Quote>,
+    next: NextFunction
+  ) => Promise<void>
+  getRates: (
+    req: Request,
+    res: Response<Rates>,
+    next: NextFunction
+  ) => Promise<void>
```
raducristianpopa, line 17: We can use the `ControllerFunction` type here.

packages/wallet-backend-update/src/rafiki/rafiki-client.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    return response.asset as Asset
+  }
+
+  public async createIncomingPayment(
+    paymentPointerId: string,
+    amount: bigint | null,
+    asset: Asset,
+    description?: string,
+    expiresAt?: string
```
raducristianpopa, line 96: Here we can use an object for the arguments.

packages/wallet-backend-update/src/rafiki/rafiki-client.ts
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    return response.createPaymentPointer.paymentPointer
+  }
+
+  public async createQuote(
+    paymentPointerId: string,
+    receiver: string,
+    asset: Asset,
+    amount?: bigint
```
raducristianpopa, line 223: Ditto above.

packages/wallet-backend-update/src/rafiki/service.ts
```diff
@@ -0,0 +1,439 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
raducristianpopa, line 2:
```suggestion
```

packages/wallet-backend-update/src/rafiki/service.ts
```diff
@@ -0,0 +1,439 @@
...
+      source_ewallet: this.deps.env.RAPYD_SETTLEMENT_EWALLET
+    })
+
+    if ((transferResult as any).status?.status !== 'SUCCESS') {
+      ;(transferResult as any).status?.message &&
+        this.deps.logger.error((transferResult as any).status.message)
+      throw new Error(
+        `Unable to transfer from ${this.deps.env.RAPYD_SETTLEMENT_EWALLET} into ${receiverWalletId}`
+      )
+    }
```
raducristianpopa, line 208:
```suggestion
    if (transferResult.status?.status !== 'SUCCESS') {
      this.deps.logger.error(transferResult.status?.message ?? 'Unable to transfer funds')
      throw new Error(
        `Unable to transfer from ${this.deps.env.RAPYD_SETTLEMENT_EWALLET} into ${receiverWalletId}`
      )
    }
```

packages/wallet-backend-update/src/rafiki/service.ts
```diff
@@ -0,0 +1,439 @@
...
+    this.deps.logger.info(
+      `Succesfully transfered ${this.amountToNumber(
+        sendAmount
+      )} from ${source_ewallet} to settlement account on ${
+        EventType.OutgoingPaymentCompleted
+      }`
+    )
+  }
+
+  private async handleOutgoingPaymentFailed(wh: WebHook) {
```
raducristianpopa, line 310: I believe it would be a good approach to specify the return type of a function, even if it's `void`. In later versions, the default return type will be `undefined`.

https://devblogs.microsoft.com/typescript/announcing-typescript-5-1-rc/#easier-implicit-returns-for-undefined-returning-functions

packages/wallet-backend-update/src/rafiki/service.ts
```diff
@@ -0,0 +1,439 @@
...
+      source_ewallet
+    })
+
+    if ((transferResult as any).status?.status !== 'SUCCESS') {
+      ;(transferResult as any).status?.message &&
+        this.deps.logger.error((transferResult as any).status.message)
+      throw new Error(
+        `Unable to transfer already sent amount from ${source_ewallet} into settlement account ${this.deps.env.RAPYD_SETTLEMENT_EWALLET} on ${EventType.OutgoingPaymentFailed}`
+      )
+    }
```
raducristianpopa, line 361: Ditto above.

packages/wallet-backend-update/src/rapyd/controller.ts
```diff
@@ -0,0 +1,159 @@
...
+import { RapydService } from './service'
+import { kycSchema, profileSchema, walletSchema } from './validation'
+
+interface IRapydController {
+  getCountryNames: ControllerFunction<{ label?: string; value?: string }[]>
+  getDocumentTypes: ControllerFunction<RapydDocumentType[]>
+  createWallet: ControllerFunction<RapydWallet>
+  verifyIdentity: ControllerFunction<RapydIdentityResponse>
+  updateProfile: ControllerFunction<RapydWallet>
+}
```
raducristianpopa, line 13:
```suggestion
type Option = {
  label?: string
  value?: string
}

interface IRapydController {
  getCountryNames: ControllerFunction<Option[]>
  getDocumentTypes: ControllerFunction<RapydDocumentType[]>
  createWallet: ControllerFunction<RapydWallet>
  verifyIdentity: ControllerFunction<RapydIdentityResponse>
  updateProfile: ControllerFunction<RapydWallet>
}
```

packages/wallet-backend-update/src/rapyd/rapyd-client.ts
```diff
@@ -0,0 +1,248 @@
+import axios, { AxiosError, AxiosResponse } from 'axios'
+import crypto from 'crypto-js'
+import { Logger } from 'winston'
+import { Env } from '@/config/env'
+
+interface IRapydClient {
+  createWallet(
+    wallet: RapydWallet
+  ): Promise<AxiosResponse<RapydResponse<RapydWallet>>>
```
raducristianpopa, line 9: Can `AxiosResponse` be dropped? In the `request` function we directly return the data.

packages/wallet-backend-update/src/rapyd/rapyd-client.ts
```diff
@@ -0,0 +1,248 @@
...
+    return this.request('put', url, body)
+  }
+
+  private calcSignature(
+    httpMethod: string,
+    url: string,
+    salt: string,
+    accessKey: string,
+    secretKey: string,
+    body: string
```
raducristianpopa, line 188: We can use an object here as well.

packages/wallet-backend-update/src/rapyd/service.ts
```diff
@@ -0,0 +1,196 @@
...
+  address: string
+  city: string
+  country: string
+  zip: string
+  email: string
+  id: string
+  phone?: string
+}
+
+export class RapydService {
```
raducristianpopa, line 35: We should create a Rapyd interface.

```suggestion
export class RapydService implements IRapydService {
```

packages/wallet-backend-update/src/rapyd/rapyd-client.ts
```diff
@@ -0,0 +1,248 @@
...
+  env: Env
+}
+
+export class RapydClient implements IRapydClient {
+  constructor(private deps: RapydClientDependencies) {}
+
+  public createWallet(
+    wallet: RapydWallet
+  ): Promise<AxiosResponse<RapydResponse<RapydWallet>>> {
+    return this.post('user', JSON.stringify(wallet))
```
raducristianpopa, line 48: I think `post`, `put` should accept a generic. The return type is still any in this case.

packages/wallet-backend-update/src/rapyd/rapyd.d.ts
beniaminmunteanu: tbd in future PR, will open issue

packages/wallet-backend-update/src/rafiki/service.ts
beniaminmunteanu: Will keep in mind.
Also 
@rico191013  @dragosp1011

packages/wallet-backend-update/src/app.ts
```diff
@@ -71,7 +102,28 @@ export class App {
...
+    const outgoingPaymentController = await this.container.resolve(
+      'outgoingPaymentController'
+    )
+    const rapydController = await this.container.resolve('rapydController')
+    const assetController = await this.container.resolve('assetController')
+    const accountController = await this.container.resolve('accountController')
+
+    app.use(
+      cors({
+        origin: 'http://localhost:4003',
```
dragosp1011, line 123: I think these settings were changes with the last Rico's PR, will be good to have them here as well

packages/wallet-backend-update/src/rafiki/rafiki-client.ts
```diff
@@ -0,0 +1,259 @@
...
+
+    return paymentResponse.payment
+  }
+
+  public async createRafikiPaymentPointer(
+    publicName: string,
+    assetId: string,
+    url: string
+    //! build the url where this is called
+    //`${env.OPEN_PAYMENTS_HOST}/${paymentPointerName}`
```
dragosp1011, line 207: this comment can be removed

packages/wallet-backend-update/src/app.ts
beniaminmunteanu: gr8,thx

package.json
```diff
@@ -18,9 +18,9 @@
     "prod": "pnpm docker:compose:prod up -d --build",
     "dev:update": "pnpm docker:compose up -d --build && pnpm wallet-frontend dev",
     "dev:backend:legacy": "pnpm wallet-backend build && pnpm dev && nodemon --watch packages/wallet-backend -e ts --exec 'pnpm docker:compose:legacy up -d --no-deps --build wallet-backend'",
+    "dev:backend:update": "pnpm wallet-backend-update build && pnpm dev:update && nodemon --watch packages/wallet-backend-update -e ts --exec 'pnpm docker:compose up -d --no-deps --build wallet-backend-update'",
     "docker:compose": "docker compose -f ./docker/dev/docker-compose.yml",
     "docker:down": "pnpm docker:compose down",
-    "docker:compose:legacy": "docker compose -f ./docker/dev/docker-compose-legacy.yml",
```
Tymmmy, line 23: If this is missing, then pnpm dev will give error:
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command failed with ENOENT: docker:compose:legacy up -d --build
spawn docker:compose:legacy ENOENT
Command "docker:compose:legacy" not found. Did you mean "pnpm run docker:down:legacy"?

packages/wallet-frontend/src/lib/api/user.ts
```diff
@@ -135,7 +135,7 @@ const createUserService = (): UserService => ({
   async signUp(args): Promise<SignUpResponse> {
     try {
       const response = await httpClient
-        .post('signup', {
+        .post('auth/signup', {
```
raducristianpopa, line 138:
```suggestion
        .post('signup', {
```

packages/wallet-frontend/src/lib/api/user.ts
```diff
@@ -151,7 +151,7 @@ const createUserService = (): UserService => ({
   async login(args): Promise<LoginResponse> {
     try {
       const response = await httpClient
-        .post('login', {
+        .post('auth/login', {
```
raducristianpopa, line 154:
```suggestion
        .post('login', {
```

cryptofan1013 (approved):

rico191013 (dismissed):

packages/wallet-frontend/src/lib/api/user.ts
beniaminmunteanu: moved backend-update to /signup /login for now aswell, without the /auth

rico191013 (approved):

PR #387, opened 2023-05-29
feat: migrate frontend to new backend

raducristianpopa:
### Context
- #373 

### Changes
- Updates frontend services to use the new backend
- Small frontend/backend refactoring
- Adds missing endpoints needed by Rafiki

PR #392, opened 2023-05-30
refactor(wallet-frontend): update user service

raducristianpopa:
### Context
- #387 

### Changes
Types update and small changes.

### DIFF
--- a/packages/wallet-backend-update/src/rapyd/controller.ts
+++ b/packages/wallet-backend-update/src/rapyd/controller.ts
@@ -10,7 +10,7 @@ interface IRapydController {
   getDocumentTypes: ControllerFunction<RapydDocumentType[]>
   createWallet: ControllerFunction<RapydWallet>
   verifyIdentity: ControllerFunction<RapydIdentityResponse>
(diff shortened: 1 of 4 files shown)


packages/wallet-backend-update/src/rapyd/controller.ts
```diff
@@ -10,7 +10,7 @@ interface IRapydController {
   getDocumentTypes: ControllerFunction<RapydDocumentType[]>
   createWallet: ControllerFunction<RapydWallet>
   verifyIdentity: ControllerFunction<RapydIdentityResponse>
-  updateProfile: ControllerFunction<RapydWallet>
+  updateProfile: ControllerFunction
```
Tymmmy, line 13: we don't need a RapydWallet types response anymore?

raducristianpopa: It was returning the whole user object from Rapyd, which we do not really need.

Tymmmy (approved):
Tested it and looks to be working as intended

dragosp1011 (approved):

PR #398, opened 2023-06-01
refactor: payment pointers service

raducristianpopa:
### Context
- #387 

### Changes

### DIFF
--- a/packages/wallet-frontend/src/pages/account/[accountId].tsx
+++ b/packages/wallet-frontend/src/pages/account/[accountId].tsx
@@ -7,7 +7,6 @@ import { AppLayout } from '@/components/layouts/AppLayout'
 
 import { PaymentPointerCard } from '@/components/PaymentPointerCard'
 import { Account, accountService } from '@/lib/api/account'
-import { PaymentPointer, paymentPointerService } from '@/lib/api/paymentPointer'
 import { useDialog } from '@/lib/hooks/useDialog'
 import { NextPageWithLayout } from '@/lib/types/app'
 
@@ -22,10 +21,7 @@ import { z } from 'zod'
 
 type AccountPageProps = InferGetServerSidePropsType<typeof getServerSideProps>
 
-const AccountPage: NextPageWithLayout<AccountPageProps> = ({
-  account,
-  paymentPointers
-}) => {
+const AccountPage: NextPageWithLayout<AccountPageProps> = ({ account }) => {
   const [openDialog, closeDialog] = useDialog()
   const formattedAmount = useMemo(
     () =>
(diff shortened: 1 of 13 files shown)


raducristianpopa:
I would like to know your opinions on these approaches:
- Fetch payment pointers along side the account when performing a `get` for an account (for account page);
- Fetch transactions along side the payment pointer when performing a `get` request for a payment pointer (for transactions page)

Using these approaches will result in less API calls since we were performing two or three in some pages in order to get information about everything we needed.

Tymmmy:
I would honestly leave it like this for now, for 2 reasons, there might be a lot of transactions and do we want to fetch them all with the payment pointers list already? And second of all for now we don't have the proposed time to make this change. If we do decide to do so, we will probably put it into Tech Debt with a new story. But for now I would definetly leave it as it is.

Tymmmy (approved):
I tested PP creation, soft deletion, and get PP lists for Account, and also PP list for selection in Send /Pay. Looks ok

packages/wallet-frontend/src/pages/account/[accountId]/[paymentPointerId].tsx
```diff
@@ -103,67 +97,35 @@ export const getServerSideProps: GetServerSideProps<{
...
-        result.data.paymentPointerId,
-        ctx.req.headers.cookie
-      ),
-      paymentPointerService.getTransactions(
-        result.data.accountId,
-        result.data.paymentPointerId,
-        ctx.req.headers.cookie
-      )
-    ])
+  const [accountResponse, paymentPointerResponse] = await Promise.all([
```
Comment (dragosp1011) [inline packages/wallet-frontend/src/pages/account/[accountId]/[paymentPointerId].tsx:100]: adding payment pointers as part of account response it is useful in some cases(you make one request instead of 2), but some times the response will have data that you don't need
for example here payment pointers are not needed in account response and if you have a lot of payment pointers that can be a significant increase in response time

Comment (dragosp1011) [reply packages/wallet-frontend/src/pages/account/[accountId]/[paymentPointerId].tsx:100]: on transactions we have order parameter in the request and we will need to add pagination and some other filters, so not sure how much sense it has to change the list of transactions request and have them included in payment pointer response instead

packages/wallet-frontend/src/pages/account/[accountId].tsx
```diff
@@ -134,27 +129,25 @@ export const getServerSideProps: GetServerSideProps<{
...
   }
 
-  if (!accountResponse.data || !paymentPointersResponse.data) {
+  if (!accountResponse.data) {
     return {
       notFound: true
     }
   }
-
+  console.log(accountResponse.data)
```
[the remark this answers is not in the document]
Comment (raducristianpopa) [reply packages/wallet-frontend/src/pages/account/[accountId].tsx:147]: Nope. Good catch!

packages/wallet-frontend/src/pages/account/[accountId]/[paymentPointerId].tsx
Comment (raducristianpopa) [reply packages/wallet-frontend/src/pages/account/[accountId]/[paymentPointerId].tsx:100]: Makes a lot of sense. Will keep things as is for the moment.

PR #408, opened 2023-06-07
fix: add missing endpoints and refactor transfer module

raducristianpopa:
### Context
- #387 

### Changes
- Adds missing modules and endpoints
- Correctly formats the amount in the transfer pages and transactions page
- Adds `codegen` config
- Fixes `createQuoteMutation`

### DIFF
--- a/packages/wallet-backend-update/migrations/20230317143023_create_transactions.js
+++ b/packages/wallet-backend-update/migrations/20230317143023_create_transactions.js
@@ -12,7 +12,7 @@ exports.up = function (knex) {
     table.foreign('paymentPointerId').references('paymentPointers.id')
 
     table.string('assetCode').notNullable()
-    table.bigint('value').notNullable()
+    table.bigint('value')
 
     table.enum('type', ['INCOMING', 'OUTGOING']).notNullable()
     table.enum('status', ['PENDING', 'COMPLETED', 'REJECTED']).notNullable()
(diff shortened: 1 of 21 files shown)


packages/wallet-backend-update/src/app.ts
```diff
@@ -133,18 +133,6 @@ export class App {
...
-    router.use('*', (req: Request, res: CustomResponse, next: NextFunction) => {
-      if (req.headers['content-type'] === 'application/json') {
-        next()
-      } else {
-        res.status(415).json({
-          success: false,
-          message: "Only 'application/json' content type is supported"
-        })
-      }
-    })
```
raducristianpopa, line 146: Rafiki was not happy with this check.

packages/wallet-backend-update/migrations/20230317143023_create_transactions.js
```diff
@@ -12,7 +12,7 @@ exports.up = function (knex) {
     table.foreign('paymentPointerId').references('paymentPointers.id')
 
     table.string('assetCode').notNullable()
-    table.bigint('value').notNullable()
+    table.bigint('value')
```
dragosp1011, line 15: why?

raducristianpopa: Line 74
https://github.com/interledger/testnet/blob/58f016239c5b518c0f835f76c70429a110fd5b44/packages/wallet-backend-update/src/outgoingPayment/service.ts#L73-L79

dragosp1011 (approved):

PR #417, opened 2023-06-12
feat: withdraw money functionality

### LINKED ISSUE #294: Withdraw money functionality
Add a Withdraw money functionality next to the Fund Account, with the icon arrow reversed.
The popup form for withdraw should be similar to fund popup form.

Also implement backend rapyd request for this and update balances.

Tymmmy:
### Context
- fixes #294 

### Changes
- Added new Withdraw button and withdraw dialog
- Added Rapyd Payout details to simulate money withdraw


packages/wallet-backend-update/src/account/service.ts
```diff
@@ -12,18 +12,21 @@ type CreateAccountArgs = {
...
 }
 
-type FundAccountArgs = {
+type FundsArgs = {
   userId: string
   amount: number
   assetCode: string
 }
+type FundAccountArgs = FundsArgs
+type WithdrawFundsArgs = FundsArgs
```
raducristianpopa, line 21:
```suggestion
type FundAccountArgs = {
  userId: string
  amount: number
  assetCode: string
}
type WithdrawFundsArgs = FundAccountArgs
```

packages/wallet-backend-update/src/account/service.ts
```diff
@@ -198,6 +201,68 @@ export class AccountService implements IAccountService {
...
+    const completePayoutResponse = await this.deps.rapyd.completePayout({
+      payout: payout.data.id,
+      amount: args.amount
+    })
+
+    if (completePayoutResponse.status.status !== 'SUCCESS') {
+      throw new Error(
+        `Unable to withdraw funds from your account: ${completePayoutResponse.status.message}`
+      )
+    }
```
raducristianpopa, line 263: Should we also create a transaction when the user withdraws money?

packages/wallet-backend-update/src/account/validation.ts
```diff
@@ -7,7 +7,7 @@ export const accountSchema = z.object({
   })
 })
 
-export const fundSchema = z.object({
+export const fundsObject = {
```
raducristianpopa, line 10: I think we can keep it as `fundSchema` and use them for both actions or just re-export `withdrawFundsSchema` without creating an object and then passing it to `zod`. E.g.:

```ts
export const fundSchema = z.object({ /* ... */ })
export const withdrawFundsSchema = fundSchema
```

packages/wallet-frontend/src/lib/api/account.ts
```diff
@@ -6,7 +6,7 @@ import {
   type SuccessResponse
 } from '../httpClient'
 
-export const fundAccountSchema = z.object({
+const accountFundsObject = {
```
raducristianpopa, line 9: Ditto above.

packages/wallet-frontend/src/components/dialogs/WithdrawFundsDialog.tsx
```diff
@@ -0,0 +1,148 @@
...
+                            onClose={closeDialog}
+                            content="Withdrawal failed. Please try again"
+                          />
+                        )
+                        return
+                      }
+
+                      if (response.success) {
+                        router.replace(router.asPath)
+                        closeDialog()
```
raducristianpopa, line 78: Having a success message would be okay I think.

packages/wallet-frontend/src/components/dialogs/WithdrawFundsDialog.tsx
```diff
@@ -0,0 +1,148 @@
...
+                      error={
+                        withdrawFundsForm.formState?.errors?.amount?.message
+                      }
+                      {...withdrawFundsForm.register('amount')}
+                    />
+                    <input
+                      type="hidden"
+                      {...withdrawFundsForm.register('assetCode')}
+                      value={account.assetCode}
+                    />
```
raducristianpopa, line 122: Is this really needed? This can be grabbed on the backend since the account has all the details about the asset.

packages/wallet-backend-update/src/account/service.ts
Tymmmy: Fund account means that I am funding the account, so that's why I wanted to have a general FUNDS account, that is being used when funding and withdrawing

packages/wallet-backend-update/src/account/service.ts
Tymmmy: The we need a transaction for funding as well, I think I will create a new stroy on this in Tech Debt, to add funding and withdrawing in transactions list

packages/wallet-frontend/src/components/dialogs/WithdrawFundsDialog.tsx
Tymmmy: Beacuse funding doesn't have success message I would leave this like this as well.

packages/wallet-frontend/src/components/dialogs/WithdrawFundsDialog.tsx
Tymmmy: Yes it is needed, because the whole account is found from the assetCode in the backend, and userID ofcourse.

packages/wallet-backend-update/src/account/service.ts
```diff
@@ -198,6 +201,68 @@ export class AccountService implements IAccountService {
     }
   }
 
+  public async withdrawFunds(args: FundAccountArgs): Promise<void> {
```
raducristianpopa, line 204:
```suggestion
  public async withdrawFunds(args: WithdrawFundsArgs): Promise<void> {
```

packages/wallet-frontend/src/components/dialogs/WithdrawFundsDialog.tsx
raducristianpopa: I've just noticed that is happening for the fund dialog as well. We are [passing the account ID](https://github.com/interledger/testnet/blob/ed1e521ec36cc7b18208a3f3b0614609a367fb7f/packages/wallet-frontend/src/lib/api/account.ts#L10) when funding/withdrawing so providing `assetCode` it's redundant. I think we should change this to:
- Only pass the `accountId` and the `amount` - we are performing a query either way to see if the account exists;
- Use the `assetCode` that is retrieved from the database query

https://github.com/interledger/testnet/blob/ed1e521ec36cc7b18208a3f3b0614609a367fb7f/packages/wallet-backend-update/src/account/service.ts#L42-L45
Should be changed to:
```ts
const existingAccount = await Account.query()
  .where('userId', args.userId)
  .where('id', args.accountId)
  .first()
```
or
```ts
const existingAccount = await Account.query().findById(args.accountId)
if(!existingAcount || existingAccount.userId !== args.userId) {
  // throw?
}
```

The backend validation schema should also be changed to accept `accountId`.

raducristianpopa (dismissed):
LGTM

packages/wallet-backend-update/src/rapyd/rapyd-client.ts
```diff
@@ -159,6 +174,29 @@ export class RapydClient implements IRapydClient {
...
+    )
+  }
+
+  public withdrawFundsFromAccount(
+    req: WithdrawFundsFromAccountRequest
+  ): Promise<RapydResponse<WithdrawFundsFromAccountResponse>> {
+    return this.post('payouts', JSON.stringify(req))
+  }
+
+  public completePayout(
```
beniaminmunteanu, line 191: so in here, I think completePayout should be private, and used inside of withdrawFunds. If the first withdraw funds call is sucessfull, then completePayout will be called further. and if that is sucessfull as well, then return.

dragosp1011 (approved):

packages/wallet-backend-update/src/account/service.ts
```diff
@@ -198,6 +201,50 @@ export class AccountService implements IAccountService {
     }
   }
 
+  public async withdrawFunds(args: WithdrawFundsArgs): Promise<void> {
+    const account = await this.findAccountById(args.accountId, args.userId)
+
+    // get list of payout method types for currency, if we get to production: get payout type required fields will be needed
+    const payoutType = await this.deps.rapyd.getPayoutMethodTypes(
```
beniaminmunteanu, line 208: I would argue that payoutType can also be private inside rapyd-client, and used by the withdrawFundsAccount, since payoutType also has no other use here in the accountService. (or does it?)

(Ditto to the completePayout made as private suggestion)

Issue #61, opened 2023-02-10
Use the Developer Keys from Settings in Open Payments


raducristianpopa:
Use actual JWkeys from Settings/Developers menu in Open Payments, to use in Payments API.
<img width="661" alt="image" src="https://user-images.githubusercontent.com/117268143/231483709-7253aaa6-fafb-4005-ac6c-d5da7a5654b3.png">

1. User generates EdDSA-Ed25519 key on their machine, or user can generate it using generateKey api on backend.
2. User Goes to the testnet developer key page, selects payment pointer
3. Uploads public key for the selected payment pointer
4. testnet takes key and generates JWK from it, to then call createPaymentPointerKey mutation to Rafiki

PR #428, opened 2023-06-14
feat(wallet/frontend): add developer keys

raducristianpopa:
### Context

- #61 

### Changes
Based on the discussions with Max, having only one key per payment pointer is our best option at the moment since we do not have a good way to determine which key corresponds to which `keyId`.


Tymmmy (approved):
Nice work!

PR #432, opened 2023-06-16
feat: developer key backend

### LINKED ISSUE #61: Use the Developer Keys from Settings in Open Payments

rico191013:
### Context

- fixes #61


packages/wallet/backend/src/account/service.ts
```diff
@@ -132,14 +129,18 @@ export class AccountService implements IAccountService {
...
+      accounts[i].balance = transformBalance(
         accountsBalance.data?.find(
-          (rapydAccount) => rapydAccount.currency === acc.assetCode
+          (rapydAccount) => rapydAccount.currency === accounts[i].assetCode
         )?.balance ?? 0,
-        acc.assetScale
+        accounts[i].assetScale
       )
-    })
+      accounts[i].paymentPointers = await PaymentPointer.query().where(
```
dragosp1011, line 139: on getting accounts the payment pointers can be received using `withGraphFetched`

also the get accounts should be moved after the check that user exists and has a wallet id (not related to your PR but it will be good to get that change as well)

dragosp1011: @raducristianpopa is the frontend expecting the list account to return the payment pointers as well?

packages/wallet/backend/src/app.ts
```diff
@@ -143,7 +143,7 @@ export class App {
     // payment pointer routes
     router.post(
       '/accounts/:accountId/payment-pointers',
-      isAuth,
```
dragosp1011, line 146: was this removed for testing purposes?

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -132,4 +134,84 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+      paymentPointerId
+    )
+
+    if (!paymentPointer) {
+      throw new NotFound()
+    }
+
+    // Check if the user owns the payment pointer.
+    // This function throws a NotFoundException.
+    await this.deps.accountService.findAccountById(
```
dragosp1011, line 152: move getting the payment pointer and validation of account into payment pointer service

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -132,4 +134,84 @@ export class PaymentPointerService implements IPaymentPointerService {
...
```
dragosp1011, line 181: same as above, define a method in payment pointer service to use it in both places

packages/wallet/backend/src/rafiki/rafiki-client.ts
```diff
@@ -78,7 +82,7 @@ export class RafikiClient implements IRafikiClient {
       CreateAssetMutation,
       CreateAssetMutationVariables
     >(createAssetMutation, { input: { code, scale } })
-
+    console.log('asset created response ---->', response)
```
dragosp1011, line 85: remove

packages/wallet/backend/src/rafiki/rafiki-client.ts
```diff
@@ -226,6 +230,33 @@ export class RafikiClient implements IRafikiClient {
...
+      }
+    })
+    console.log('response ---->', response)
+    if (!response.createPaymentPointerKey.success) {
+      throw new Error(response.createPaymentPointerKey.message)
+    }
+    if (!response.createPaymentPointerKey.paymentPointerKey) {
+      throw new Error('Unable to fetch created payment pointer key')
+    }
+    console.log(
```
dragosp1011, line 253: remove

packages/wallet/backend/src/account/service.ts
raducristianpopa: >is the frontend expecting the list account to return the payment pointers as well?

Yes, I was thinking about a separate endpoint but this works as well.

packages/wallet/backend/src/paymentPointer/validation.ts
```diff
@@ -28,3 +28,10 @@ export const paymentPointerSchema = z.object({
       .min(3, { message: 'Public name must be at least 3 characters long' })
   })
 })
+
+export const developerKeySchema = z.object({
+  body: z.object({
+    paymentPointerId: z.string().trim(),
```
raducristianpopa, line 34:
```suggestion
    paymentPointerId: z.string().trim().uuid(),
```

packages/wallet/backend/package.json
```diff
@@ -4,7 +4,8 @@
     "start": "node -r tsconfig-paths/register dist/index.js",
     "build": "tsc --build tsconfig.build.json && tsc-alias -p tsconfig.build.json",
     "test": "jest --passWithNoTests --maxWorkers=75%",
-    "generate": "graphql-codegen --config codegen.yml"
+    "generate": "graphql-codegen --config codegen.yml",
+    "server": "npm run build && npm run start"
```
raducristianpopa, line 8: Why do we need this script?

raducristianpopa (commented):
I am a bit confused regarding the `registerKey` controller. Would love to get more context. 🤔

packages/wallet/backend/src/account/service.ts
dragosp1011: @raducristianpopa as on developer keys view there is no need for balance (no need to request rapyd) and on other pages there is not need for all the payment pointers, as they are listed separately for selected account - I think it makes more sense to have a separate endpoint

packages/wallet/backend/src/account/service.ts
```diff
@@ -132,14 +129,18 @@ export class AccountService implements IAccountService {
       user.rapydWalletId
     )
 
-    accounts.forEach((acc) => {
-      acc.balance = transformBalance(
+    for (let i = 0; i < accounts.length; i++) {
```
beniaminmunteanu, line 132: is there a preference for explicit for loops?

packages/wallet/backend/src/account/service.ts
beniaminmunteanu, line 139: Just another option here, that would have bettern performance but worse maintainability IMO:

make a single initial query that gets all payment pointers of all accounts of a user (since we have the userId) -> one to many with accounts -> one to many with payment pointers. (a couple of sql joins needed)

This way you don't get the tcp overhead of every separate sql query, and also benefit on the db engine optimisations.

However, since we won't be running in any performance issues with the scope of our project, both options are valid for me.
@raducristianpopa  @dragosp1011

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -132,4 +134,83 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+      paymentPointer.accountId,
+      userId
+    )
+
+    return paymentPointer
+  }
+
+  async generateKeyPair(
```
dragosp1011, line 160: IMO will be goof to extract key generation in a separate service

packages/wallet/backend/src/app.ts
```diff
@@ -201,6 +201,18 @@ export class App {
     router.post('/quote', rafikiController.createQuote)
     router.post('/webhooks', rafikiController.onWebHook)
 
+    //payment-pointer
+    router.post(
+      '/payment-pointer/register-key',
```
dragosp1011, line 206: I think it make sense to change it to `/accounts/:accountId/payment-pointers/:id/register-key` so it aligns with other payment pointer endpoints

packages/wallet/backend/src/account/controller.ts
```diff
@@ -53,13 +52,30 @@ export class AccountController implements IAccountController {
...
+  ) => {
+    const userId = req.session.user.id
+
+    try {
+      const accounts =
+        await this.deps.accountService.getAccountsWithGraphFetched(userId)
+
+      res
+        .status(200)
+        .json({ success: true, message: 'SUCCESS', data: accounts })
```
raducristianpopa, line 78: Instead of having a separate function for this, should we have a query param (`include`) where we provide what else need to be fetched? In our case it will be something like `http://localhost/accounts?include=paymentPointers`.

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -132,4 +134,100 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+
+    return { privateKey, publicKey }
+  }
+
+  async registerKey(
+    userId: string,
+    accountId: string,
+    paymentPointerId: string
+  ): Promise<void> {
+    const paymentPointer = await this.getPaymentPointerAfterValidation(
```
raducristianpopa, line 191: I think this should be moved into the controller instead.

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -132,4 +134,100 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+      userId,
+      accountId,
+      paymentPointerId
+    )
+
+    await this.deps.accountService.findAccountById(accountId, userId)
```
dragosp1011, line 197: this is not needed as on getting the payment pointer the account is validated as well

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -132,4 +134,100 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+
+    if (!paymentPointer) {
```
dragosp1011, line 199: this check is also part of getting the payment pointer

packages/wallet/backend/src/utils/jwk.ts
```diff
@@ -0,0 +1,21 @@
+import { createPublicKey, generateKeyPairSync } from 'crypto'
+
+export const generateJWK = () => {
```
mkurapov, line 3: you could also `pnpm i @interledger/http-signature-utils` to get the `generateJwk` function here

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -132,4 +134,100 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+      throw new NotFound()
+    }
+
+    const jwkStr = paymentPointer.jwk
+    if (!jwkStr) {
+      throw new Error(`jwk doesnt' exist.`)
+    }
+
+    await this.deps.rafikiClient.createRafikiPaymentPointerKey(
+      this.generateJwk(paymentPointer.accountId, jwkStr),
```
mkurapov, line 209: Since you could have multiple keys, I think we'd need to have something other or something in addition to just the accountId

packages/wallet/backend/src/app.ts
```diff
@@ -169,6 +169,18 @@ export class App {
...
+      '/accounts/:accountId/payment-pointers/:paymentPointerId/register-key',
+      isAuth,
+      paymentPointerController.registerKey
+    )
+
+    router.post(
+      '/accounts/:accountId/payment-pointers/:paymentPointerId/generate-key',
+      isAuth,
+      paymentPointerController.generateKey
+    )
```
mkurapov, line 182: Why are both `/register-key` and `/generate-key` needed? Would you be able to do everything in a single `/generate-key` call?

packages/wallet/backend/src/utils/jwk.ts
rico191013: image

It seems not compatible...

mkurapov: I think what is not compatible here is the `rafikiClient.createRafikiPaymentPointerKey` function. You would just need to  update how you pass in the `generatedJwk`

mkurapov: If you don't want to go that route, we should remove "@interledger/http-signature-utils" in that case

packages/wallet/backend/migrations/20230620082429_add_publicKey_to_payment_pointers.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.alterTable('paymentPointers', (table) => {
+    table.jsonb('keyIds')
```
mkurapov, line 7: As an alternative, this could be an array: packages/wallet/backend/migrations/20230620082429_add_publicKey_to_payment_pointers.js

packages/wallet/backend/src/paymentPointer/controller.ts
```diff
@@ -109,4 +114,30 @@ export class PaymentPointerController implements IPaymentPointerController {
...
+        await this.deps.paymentPointerService.registerKey(
+          userId,
+          accountId,
+          paymentPointerId
+        )
+
+      res.status(200).json({
+        success: true,
+        message: 'Public key is successfully registered',
+        data: { privateKey, publicKey }
```
mkurapov, line 137: Maybe keyId might be useful to return as well? Otherwise, if the user can only see a list of keyids, without knowing when they were created, they wouldn't know which id corresponds to which key

packages/wallet/backend/migrations/20230620082429_add_publicKey_to_payment_pointers.js
rico191013: Well, I think only jsonb is supported in knex.

dragosp1011 (approved):

packages/wallet/backend/src/account/controller.ts
```diff
@@ -51,15 +50,19 @@ export class AccountController implements IAccountController {
...
   ) => {
     const userId = req.session.user.id
+    const shouldGraphFetched = req.query['include'] == 'paymentPointers'
 
     try {
-      const getAccountsResult = await this.deps.accountService.getAccounts(
-        userId
-      )
+      let accounts
+      if (shouldGraphFetched)
```
casalete, line 57: I think you should pass 'paymentPointers' as optional args to the already existing getAccounts method, in order to query with graph fetched instead of adding the 'getAccountsWithGraphFetched' method

packages/wallet/backend/src/account/service.ts
```diff
@@ -145,6 +141,20 @@ export class AccountService implements IAccountService {
     return accounts
   }
 
+  public async getAccountsWithGraphFetched(userId: string): Promise<Account[]> {
```
casalete, line 144: this should be an extension of the already existing getAccounts method.

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -132,4 +135,41 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+
+    const { privateKey, publicKey } = generateKeyPairSync('ed25519')
+    const publicKeyPEM = publicKey
+      .export({ type: 'spki', format: 'pem' })
+      .toString()
+    const privateKeyPEM = privateKey
+      .export({ type: 'pkcs8', format: 'pem' })
+      .toString()
+    const keyId = uuid()
+    console.log('-----')
```
casalete, line 158: are these logs needed? or just leftovers from development

packages/wallet/backend/src/account/controller.ts
dragosp1011: yeap, I agree with this one @

rico191013: Well, in that case getAccounts function will get complex again, and it's same as previous version. But @dragosp1011 and @raducristianpopa wanted it to be separated...

dragosp1011: I think something like this should work
```
const user = await User.query().findById(userId)

    if (!user || !user.rapydWalletId) {
      throw new NotFound()
    }

    const query = Account.query().where('userId', userId)

    if (paymentPointers) {
      query.withGraphFetched('paymentPointers')
    }

    const accounts = await query

    if (balance) {
      const accountsBalance = await this.deps.rapyd.getAccountsBalance(
          user.rapydWalletId
      )

      accounts.forEach((acc) => {
        acc.balance = transformBalance(
            accountsBalance.data?.find(
                (rapydAccount) => rapydAccount.currency === acc.assetCode
            )?.balance ?? 0,
            acc.assetScale
        )
      })
    }

    return accounts
    ```

dragosp1011: @raducristianpopa what do you think? is this what you had in mind?

dragosp1011 (approved):

PR #440, opened 2023-06-23
feat(backend): Rapyd future-proof

### LINKED ISSUE #378: Rapyd return types as zod schemas
rapyd.d.ts types should be changed to be zod schemas, and then the types infered from the zod schemas and exported furhter.

This is in order to catch potential future changes that rapyd does to their api.

beniaminmunteanu:
-closes #378 

### Context

We are using Rapyd Sandbox and we are accessing it through an URL. This means we cannot stale the version of rapyd. As such, if any changes happen to Rapyd API, we might be affected.

In order to mitigate the effect of rapyd changes, Testnet will now validate the rapyd response against defined ZOD schemas, for easier debugging and further fixes if needed.

### Changes

- Zod schemas defined for any rapyd interaction
- Types are infered from zod schemas
- Rapyd client now applies zod schemas on  the responses received from rapyd


packages/wallet/backend/src/rapyd/rapyd-client.ts
```diff
@@ -61,86 +102,95 @@ type CalculateSignatureParams = {
...
 
-  public getDocumentTypes(
+  public async getDocumentTypes(
     country: string
   ): Promise<RapydResponse<RapydDocumentType[]>> {
-    return this.get<RapydResponse<RapydDocumentType[]>>(
+    return this.get(
       `identities/types?country=${country}`
-    )
+      // RapydDocumentsTypeSchema
```
beniaminmunteanu, line 162: I am having trouble in validating and infering types in cases where we are expecting arrays as a result from rapyd.

For now, all these cases just don't have validation applied, and I made the private request method also be able to just blindly return responses without any validation or type inference.

Any suggestions or contributions here are welcome

packages/wallet/backend/src/rapyd/validation.ts
```diff
@@ -36,3 +36,18 @@ export const walletSchema = z.object({
     phone: z.string().optional()
   })
 })
+
+export async function validateRapydResponse<
```
beniaminmunteanu, line 40: I preffered to break the DRY principle here and just copy most of this global method that I previously built for validation of our frontend client requests.

packages/wallet/backend/src/rapyd/rapyd-client.ts
```diff
@@ -328,19 +391,30 @@ export class RapydClient implements IRapydClient {
...
         method,
         url: `${this.deps.env.RAPYD_API}/${url}`,
         ...(body && { data: body }),
         headers
       })
+
+      if (zodResponseSchema) {
+        await validateRapydResponse(
+          zodResponseSchema as AnyZodObject | ZodEffects<AnyZodObject>,
+          res.data
```
dragosp1011, line 414: I think here should be `res.data.data` as one `data` is from axios and another one is rapyd response

packages/wallet/backend/src/rapyd/rapyd-client.ts
dragosp1011: maybe make every schema to have the `data` and `status` field that rapyd returns so every response will be an object

beniaminmunteanu: So you suggest I give up the following structure that I am currently using, in which RapydResponse has a data object of type *whatever rapyd response type*:

`export interface RapydResponse<T> {
  status: z.TypeOf<typeof StatusSchema>
  data: T
}`

used like:

`RapydResponse<RapydDepositResponse>`


and just add the status + data items in every rapyd response type / schema ?

dragosp1011: yes, as an workaround to not deal with array response type

rico191013 (approved):

packages/wallet/backend/src/rapyd/rapyd-client.ts
beniaminmunteanu: @raducristianpopa  @rico191013  any other thoughts here? I'm still hoping to get the types working with the current structure which implies the generics used in RapydResponse:
`export interface RapydResponse<T> { status: z.TypeOf<typeof StatusSchema> data: T }`

If not, I will prolly go for what Dragos suggested.

PR #445, opened 2023-06-23
fix: Move X from PPs after View button

### LINKED ISSUE #444: PP list - X for deletion, move it next to View
We should move the red X for the deletion next to the View button. Probably after it.

Tymmmy:
### Context

- fixes #444 

### Changes
- received from testnet feedback

### DIFF
--- a/packages/wallet/frontend/src/components/PaymentPointerCard.tsx
+++ b/packages/wallet/frontend/src/components/PaymentPointerCard.tsx
@@ -67,9 +67,20 @@ export const PaymentPointerCard = ({
 
   return (
     <div ref={cardRef} className={cardStyles}>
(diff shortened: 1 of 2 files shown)


packages/wallet/frontend/src/components/PaymentPointerCard.tsx
```diff
@@ -67,9 +67,20 @@ export const PaymentPointerCard = ({
 
   return (
     <div ref={cardRef} className={cardStyles}>
+      <div className="flex flex-1 items-center justify-between space-x-2">
+        <>
```
raducristianpopa, line 71: This fragment can be removed.

raducristianpopa (approved):

PR #449, opened 2023-06-26
feat(transactions): add fund/withdraw transactions

### LINKED ISSUE #422: Funds & Withdraw should appear in transactions list
When user funds or withdraws money, this should appear as a transaction in the transctions list

dragosp1011:
### Context

- fixes #422 

### Changes
 - Alters transactions table to support saving account id, payment id to be a string and payment pointer to be nullable.
 - Registers transactions for withdraw/fund account operations.
 
 The transactions will not be displayed on frontend as the transaction table is for a payment pointer at the moment.

### DIFF
--- a/packages/wallet/backend/src/account/service.ts
+++ b/packages/wallet/backend/src/account/service.ts
@@ -5,6 +5,7 @@ import { RapydClient } from '@/rapyd/rapyd-client'
 import { Logger } from 'winston'
 import { RafikiClient } from '@/rafiki/rafiki-client'
 import { transformBalance } from '@/utils/helpers'
+import { Transaction } from '@/transaction/model'
 
 type CreateAccountArgs = {
   userId: string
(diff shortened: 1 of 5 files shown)


packages/wallet/backend/src/account/service.ts
```diff
@@ -221,6 +236,28 @@ export class AccountService implements IAccountService {
...
+        args.amount,
+        account.assetId
+      ),
+      type: 'OUTGOING',
+      status: 'COMPLETED',
+      description: 'Withdraw funds'
+    })
+  }
+
+  async transformAmountBasedOnAsset(
```
raducristianpopa, line 254: https://github.com/interledger/testnet/blob/bd82d141d1c0148b289d8a3aa896b35690331223/packages/wallet/backend/src/utils/helpers.ts#L14-L16

I think this can be used instead of having another method in the service. 🤔 Maybe `transformBalance` is not a good variable name in this use case.

Tymmmy (approved):

PR #454, opened 2023-06-28
feat(wallet): use payment pointer format instead of `https://`

### LINKED ISSUE #452: Allow users to use the correct format for payment pointers
At the moment the payment pointers can only be used as an URL. The users should be allowed to use the payment pointer format that is described in the [spec](https://paymentpointers.org/#explainer).

E.g.: When sending money, the user is required to specify the receiver's payment pointer (`https://ilp.rafiki.money/bob`). They should be allowed to use this format: `$ilp.rafiki.money/bob`.

raducristianpopa:
### Context

- fixes #452


Tymmmy (changes requested):
Change for ILP PP Onboarding

Tymmmy (approved):

Tymmmy (approved):

dragosp1011 (approved):

PR #455, opened 2023-06-28
feat(wallet/backend): add default account and payment pointer for a new user

### LINKED ISSUE #362: Add default account & payment pointer & funds
When the user is creating a new account, they will get faced with an empty home page.
Let's create a default EUR account (My EUR Account or UserFirstName's EUR Account), with a default payment pointer (eur_pp), and with default funds (10000 EUR). We will use EUR, because we will do an Onboarding Flow where user will create (will be forced to create) an USD account.

raducristianpopa:
### Context

- fixes #362 

### Changes


Tymmmy (approved):

packages/wallet/backend/src/rapyd/controller.ts
```diff
@@ -81,6 +90,36 @@ export class RapydController implements IRapydController {
       req.session.user.needsWallet = false
       await req.session.save()
 
+      if (this.deps.env.NODE_ENV === 'production') {
+        const asset = (await this.deps.rafikiClient.listAssets()).find(
```
dragosp1011, line 94: I suggest to extract this into a `createDefaultAccount` method

raducristianpopa: Having it inside the `rapydService` wouldn't make sense. I abstracted away the `createDefaultAccount` method. The payment pointer creation does not happen in this method because we will end up with circular dependencies.

packages/wallet/backend/src/rapyd/controller.ts
```diff
@@ -81,6 +86,21 @@ export class RapydController implements IRapydController {
       req.session.user.needsWallet = false
       await req.session.save()
 
+      const defaultAccount =
+        await this.deps.accountService.createDefaultAccount(id)
+      if (defaultAccount) {
+        const typedArray = new Uint32Array(1)
```
dragosp1011, line 92: I think create default payment pointer also can be a separate method in payment pointer service

raducristianpopa: I don't think it requires a different method in the payment pointer service.

Tymmmy (approved):

PR #456, opened 2023-06-28
feat: Forgot password functionality

### LINKED ISSUE #299: Forgotten password functionality
At the login page implement Forgot password functionality.
Best practice would be to send a reset password link for the user to pick a new password, and not resend password in the email.

Tymmmy:
### Context

- fixes #299


docker/dev/docker-compose-prod.yml
```diff
@@ -57,6 +57,8 @@ services:
       OPEN_PAYMENTS_HOST: https://ilp.rafiki.money
       GRAPHQL_ENDPOINT: http://rafiki-backend:3001/graphql
       RAFIKI_MONEY_FRONTEND_HOST: rafiki.money
+      SENDGRID_APFROM_EMAILI_KEY: ${SENDGRID_API_KEY}
```
raducristianpopa, line 60:
```suggestion
      SENDGRID_API_KEY: ${SENDGRID_API_KEY}
```

packages/wallet/backend/src/app.ts
```diff
@@ -147,6 +149,8 @@ export class App {
     // Auth Routes
     router.post('/signup', authController.signUp)
     router.post('/login', authController.logIn)
+    router.post('/reset-password', userController.requestResetPassword)
```
raducristianpopa, line 152:
```suggestion
    router.post('/forgot', userController.requestResetPassword)
```

packages/wallet/backend/src/email/service.ts
```diff
@@ -0,0 +1,43 @@
...
+  private readonly baseUrl: string
+  constructor(private deps: EmailServiceDependencies) {
+    sendgrid.setApiKey(this.deps.env.SENDGRID_API_KEY)
+    this.from = this.deps.env.FROM_EMAIL
+    const host = this.deps.env.RAFIKI_MONEY_FRONTEND_HOST
+    this.baseUrl =
+      host === 'localhost' ? 'http://localhost:4003' : `https://${host}`
+  }
+
+  async send(email: EmailArgs): Promise<void> {
```
raducristianpopa, line 31: `send` can be a `private` function.

packages/wallet/backend/src/email/service.ts
```diff
@@ -0,0 +1,43 @@
...
+    await sendgrid.send(email)
+  }
+
+  async sendForgotPassword(to: string, token: string): Promise<void> {
+    return this.send({
+      from: this.from,
+      to,
+      subject: 'Forgotten Password - rafiki.money',
```
raducristianpopa, line 39:
```suggestion
      [RAFIKI MONEY] Reset your password',
```

packages/wallet/backend/src/user/model.ts
```diff
@@ -1,15 +1,15 @@
...
+import { Model, ModelOptions, QueryContext } from 'objection'
+import { bcryptVerify } from 'hash-wasm'
+import { encryptPassword } from '@/utils/helpers'
 
 export class User extends BaseModel {
   static tableName = 'users'
 
   public email!: string
-  private password!: string
+  public password!: string
```
raducristianpopa, line 12: I think we should find a way to keep the password `private` or exclude it from every response.

packages/wallet/frontend/src/pages/auth/reset/[token].tsx
```diff
@@ -0,0 +1,157 @@
...
+        width={400}
+        height={200}
+      />
+      <p className="mt-auto font-extralight text-green">
+        Remembered your credentials?{' '}
+        <Link href="/auth/login" className="font-medium underline">
+          Login
+        </Link>
+      </p>
+    </>
```
Comment (raducristianpopa) [inline packages/wallet/frontend/src/pages/auth/reset/[token].tsx:127]: I would probably go for a `Cancel` button.

packages/wallet/frontend/src/pages/auth/reset/[token].tsx
```diff
@@ -0,0 +1,157 @@
...
+}
+
+const querySchema = z.object({
+  token: z.string()
+})
+
+export const getServerSideProps: GetServerSideProps<{
+  token: string
+}> = async (ctx) => {
+  const result = querySchema.safeParse(ctx.query)
```
Comment (raducristianpopa) [inline packages/wallet/frontend/src/pages/auth/reset/[token].tsx:138]: We should have a `check-token` controller to verify if the token exists or not. At the moment, any token will render the reset password page.

packages/wallet/frontend/src/pages/auth/forgot.tsx
```diff
@@ -0,0 +1,95 @@
...
+        width={400}
+        height={200}
+      />
+      <p className="mt-auto font-extralight text-green">
+        Remembered your credentials?{' '}
+        <Link href="login" className="font-medium underline">
+          Login
+        </Link>
+      </p>
+    </>
```
raducristianpopa, line 87: Ditto. I would go for a `Cancel` button.

beniaminmunteanu (dismissed):

packages/wallet/backend/src/email/service.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  async sendForgotPassword(to: string, token: string): Promise<void> {
+    return this.send({
+      from: this.from,
+      to,
+      subject: '[Rafiki.Money] Reset your password',
+      html: getForgotPasswordEmail(this.baseUrl, token)
+    })
+  }
```
raducristianpopa, line 41: For the development environment we should log the URL instead of sending the mail.

dragosp1011: added an env variable that allows to enable/disable send email, by default it is disabled

docker/dev/.env.example
```diff
@@ -11,3 +11,4 @@ JWT_REFRESH_TOKEN_SECRET=
 RAFIKI_STREAM_SECRET=
 SENDGRID_API_KEY=
 FROM_EMAIL=
+SEND_EMAIL=
```
raducristianpopa, line 14: Another environment variables is probably not needed. We can just check the `NODE_ENV`.

dragosp1011: a new env variable allows to enable send email on dev for testing purposes without affecting anything else

Tymmmy: Yes, we sometimes do need to send email in dev environment as well, to test somethings, so maybe an env variable is the best for now

packages/wallet/backend/src/email/service.ts
```diff
@@ -0,0 +1,54 @@
...
+interface IEmailService {
+  sendForgotPassword(to: string, token: string): Promise<void>
+}
+interface EmailServiceDependencies {
+  env: Env
+  logger: Logger
+}
+
+export class EmailService implements IEmailService {
+  private readonly from: string
```
raducristianpopa, line 22:
```suggestion
```
`this.deps.env.FROM_EMAIL` can be used directly.

packages/wallet/backend/src/email/service.ts
```diff
@@ -0,0 +1,54 @@
...
+  private readonly baseUrl: string
+  private readonly isSendEmailEnabled: boolean
```
raducristianpopa, line 24:
```suggestion
```
Ditto above. `this.deps.env.SEND_EMAIL` can be used directly.

packages/wallet/backend/src/email/service.ts
```diff
@@ -0,0 +1,54 @@
...
+      ? await this.send({
+          from: this.from,
+          to,
+          subject: '[Rafiki.Money] Reset your password',
+          html: getForgotPasswordEmail(url)
+        })
+      : this.deps.logger.info(
+          `Send email is disabled. Reset password link is: ${url}`
+        )
+  }
```
raducristianpopa, line 53: Better readability using an `if` statement.

raducristianpopa (commented):
I maintain my stance that an environment variable is unnecessary to control either or not an should be sent. We already have a lot of environment variables. To test the templates, a straightforward script that sends an email would be a perfect fit.

docker/dev/.env.example
dragosp1011: @raducristianpopa 
> I maintain my stance that an environment variable is unnecessary to control either or not an email should be sent. We already have a lot of environment variables. To test the templates, a straightforward script that sends an email would be a perfect fit.

I don't see any issues with having a lot of env variables - for sure we will have more.
Testing the template with a separate script will require additional time from the engineer to prepare it.
If an email is not sent in the testing process, how to ensure in development that send email functionality works properly and there are no issues?

Tymmmy: SO let's agree to disagree for now on this, and leave it in .env :)

beniaminmunteanu: @raducristianpopa  I get your point as well, but I am voting for the env variable for this. I think it's not that much of an overhead in configuring environments of our application, and I think it might be better than having the process context itself as the deciding factor.

beniaminmunteanu: Unresolved this for @raducristianpopa  to see

casalete (approved):

beniaminmunteanu (approved):

PR #466, opened 2023-07-04
feat(wallet/backend): add revoke payment pointer key

### LINKED ISSUE #458: Revoke payment pointer key
Using the `revokePaymentPointerKey` mutation withing the GraphQL API.

raducristianpopa:
### Context

- fixes #458

### DIFF
--- a/packages/wallet/backend/src/app.ts
+++ b/packages/wallet/backend/src/app.ts
@@ -185,6 +185,9 @@ export class App {
       isAuth,
       paymentPointerController.registerKey
     )
+    router.patch('/accounts/:accountId:/payment-pointers/:paymentPointerId/revoke-key',
+                 isAuth,
+                 paymentPointerController.revokeKey)
 
     // incoming payment routes
     router.post('/incoming-payments', isAuth, incomingPaymentController.create)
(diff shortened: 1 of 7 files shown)


packages/wallet/frontend/src/lib/api/paymentPointer.ts
```diff
@@ -207,6 +216,22 @@ const createPaymentPointerService = (): PaymentPointerService => ({
...
+      const response = await httpClient
+        .patch(
+          `accounts/${args.accountId}/payment-pointers/${args.paymentPointerId}/revoke-key`
+        )
+        .json<SuccessResponse>()
+      return response
+    } catch (error) {
+      return getError(
+        error,
+        'We were not able to revoke the key. Please try again'
```
Tymmmy, line 232:
```suggestion
        'We were not able to revoke the key. Please try again.'
```

Tymmmy (approved):

packages/wallet/backend/src/app.ts
```diff
@@ -185,6 +185,11 @@ export class App {
       isAuth,
       paymentPointerController.registerKey
     )
+    router.patch(
```
dragosp1011, line 188: maybe use delete, and instead of revoke-key -> keys

raducristianpopa: At the moment we only support one key per payment pointer, that's why I went for `PATCH` and `revoke-key`.

dragosp1011 (approved):

PR #492, opened 2023-07-11
feat(wallet): pp edit public name and allow hyphens

### LINKED ISSUE #473: Payment pointer further enhancement
From feedback users would like:

- To name PPs also using '-', not just '_'
- To edit public name of PP

raducristianpopa:
### Context

- closes #473 

### Changes
- the `X` icon was removed from the payment pointer card. Now we have a small pop-up menu that opens up when the three dot icon is clicked. The menu offers the ability to edit the payment pointer (only the public name) and to delete it;
- on smaller screens the menu opens at the bottom of the page;
- hyphens are now allowed in a payment pointer's name;


packages/wallet/frontend/src/components/cards/PaymentPointerCard.tsx
```diff
@@ -49,74 +52,172 @@ export const PaymentPointerCard = ({
...
+                          onClose={closeDialog}
+                        />
+                      )
+                    }
+                  />
+                  <Option
+                    name="Delete"
+                    intent="danger"
+                    Icon={Trash}
+                    aria-label="delete payment paymenter"
```
Tymmmy, line 155: delete what what? :D

Tymmmy (approved):

dragosp1011 (approved):

PR #500, opened 2023-07-12
feat(wallet): developer keys improvements

### LINKED ISSUE #467: API keys further echancement
In the Developers API keys we should
- download the PRIVATE KEY into a private.key (or other name) directly after Generate button click
- we should rename Generate button to Generate Private Key
- for every PP we should display Key ID + Copy button, and Last modified info
- make sure after Revoke Key is done, we can regenerate Private key

raducristianpopa:
### Context

- closes #467 

### Changes
- When the public and private key generate, we automatically save the the private key on the user's device.
- Components refactoring
- Added `publicKey` and `createdOn` fields to `keyIds`
- We are now saving and showing the public key as well


packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -200,7 +201,23 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+
+    if (!paymentPointer.keyIds) {
+      return
+    }
+
+    const trx = await PaymentPointer.startTransaction()
+    try {
+      await Promise.all([
+        await paymentPointer.$query(trx).patch({ keyIds: null }),
+        await this.deps.rafikiClient.revokePaymenterPointerKey(
```
Tymmmy, line 213: what is paymenter? I've seen it before in the code, Should it be paymentPointer?

packages/wallet/backend/src/rafiki/rafiki-client.ts
```diff
@@ -278,6 +283,19 @@ export class RafikiClient implements IRafikiClient {
     return response.createPaymentPointerKey.paymentPointerKey
   }
 
+  public async revokePaymenterPointerKey(id: string): Promise<void> {
```
Tymmmy, line 286: paymenter?

packages/wallet/frontend/src/components/settings/DeveloperKeys.tsx
```diff
@@ -0,0 +1,344 @@
...
+      )
+      return
+    }
+
+    if (response.data) {
+      const { privateKey } = response.data
+
+      generateAndDownloadFile({
+        content: privateKey,
+        filename: 'private.key',
```
Tymmmy, line 264: can it be fileName?

Tymmmy (approved):

PR #503, opened 2023-07-14
feat: verify email fror new user

### LINKED ISSUE #300: Verify Email
When registering, an email should be sent to the user with a link to verify email.

dragosp1011:
### Context

- fixes #300


packages/wallet/frontend/src/pages/auth/signup.tsx
```diff
@@ -32,7 +32,7 @@ const SignUpPage: NextPageWithLayout = () => {
               openDialog(
                 <SuccessDialog
                   onClose={closeDialog}
-                  content="Your account was created."
+                  content="Welcome to Interledger Testnet. Please verify your email address."
```
raducristianpopa, line 35: We should provide clearer instructions about the next steps for the user. It might be helpful to display a message indicating that an email has been sent to verify the email address and guide them on what to do next.

packages/wallet/backend/src/user/controller.ts
```diff
@@ -133,4 +134,24 @@ export class UserController implements IUserController {
       next(e)
     }
   }
+
+  verifyEmail = async (
```
raducristianpopa, line 138: I think this should be moved into the auth module.

raducristianpopa (commented):
To enhance the sign-up process, we should consider checking the MX records for the email domain. This will allow us to verify if we are being routed to a legitimate mail server before sending the sign-up email. Additionally, it would be beneficial filter the given email within a list of domains for disposable email services to ensure that such emails are not accepted alongside a rate limiting strategy that will help us to prevent spamming.

packages/wallet/backend/src/email/templates/images/Verify.png
rico191013, line 1: well, I think this image need to be in different folder?

Tymmmy: Hmmm well, because of the email html templates are in the backend, we decided with dragos to put them next to where they are being used in the email template. It would not make more sense to move them to frontend somewhere, because the emails are here in backend...

packages/wallet/backend/src/user/service.ts
```diff
@@ -31,7 +33,23 @@ export class UserService implements IUserService {
...
+    })
+
+    await this.deps.emailService
+      .sendVerifyEmail(args.email, token)
+      .catch((e) => {
+        this.deps.logger.error(
+          `Error on sending verify email for user ${user.email}`,
+          e
+        )
+      })
```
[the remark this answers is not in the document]
beniaminmunteanu: I'd agree on the token generation, but for me the email sending and updating of the user entities in the database could also sit here.

packages/wallet/backend/src/user/controller.ts
dragosp1011: moved controller method, but the service method remains in user service as all the work is with User model.

packages/wallet/backend/src/user/service.ts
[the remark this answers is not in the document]
dragosp1011: moved the logic into a signup method in auth service

Tymmmy (approved):

packages/wallet/backend/src/auth/service.ts
```diff
@@ -4,27 +4,55 @@ import addSeconds from 'date-fns/addSeconds'
...
+import { getRandomToken, hashToken } from '@/utils/helpers'
+import { EmailService } from '@/email/service'
+import { Logger } from 'winston'
 
 interface AuthorizeArgs {
   email: string
   password: string
 }
+
+interface SignupArgs extends AuthorizeArgs {}
```
raducristianpopa, line 16:
```suggestion
interface SignUpArgs extends AuthorizeArgs {}
```

packages/wallet/backend/src/auth/service.ts
```diff
@@ -4,27 +4,55 @@ import addSeconds from 'date-fns/addSeconds'
...
+
 interface AuthorizeResult {
   user: User
   session: Session
 }
 
 interface IAuthService {
   authorize(args: AuthorizeArgs): Promise<AuthorizeResult>
+  signup(args: SignupArgs): Promise<User>
```
raducristianpopa, line 25:
```suggestion
  signUp(args: SignUpArgs): Promise<User>
```

packages/wallet/backend/migrations/20230711171065_add_password_reset_fields.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.alterTable('users', (table) => {
+    table.boolean('isEmailVerified').defaultTo(false)
```
raducristianpopa, line 7: Should we handle the use-case in which we already have users that are registered and set the `isEmailVerified` column to `true` ?

packages/wallet/backend/src/auth/service.ts
```diff
@@ -4,27 +4,55 @@ import addSeconds from 'date-fns/addSeconds'
...
+      password,
+      verifyEmailToken: hashToken(token)
+    })
+
+    await this.deps.emailService.sendVerifyEmail(email, token).catch((e) => {
+      this.deps.logger.error(
+        `Error on sending verify email for user ${user.email}`,
+        e
+      )
+    })
```
raducristianpopa, line 51: I don't think that's the correct flow. We should:
1. do the email verification - we do not create the user if the email checks do not pass
2. create the user
3. send the email

packages/wallet/backend/src/email/service.ts
```diff
@@ -49,4 +55,23 @@ export class EmailService implements IEmailService {
...
+    const domain = to.split('@')[1]
+    await verifyDomain(domain)
+
+    const url = `${this.baseUrl}/auth/verify/${token}`
+
+    if (this.deps.env.SEND_EMAIL) {
+      return this.send({
+        to,
+        subject: '[Rafiki.Money] Verify your account',
+        html: getVerifyEmail(url)
```
raducristianpopa, line 69: I thought this was a function that check the email again. We should name this accordingly - that it retrieves the email template for a verification email.

packages/wallet/backend/src/email/verify-domain.ts
raducristianpopa, line 1: We can easily move this into the email service.

packages/wallet/backend/migrations/20230711171065_add_password_reset_fields.js
dragosp1011: I thought that we will start from scratch and it is not needed, but it won't hurt doing it

Tymmmy (approved):

beniaminmunteanu (approved):
LGTM 🚀

raducristianpopa (approved):

Issue #259, opened 2023-04-17
Add new ALL Transactions list, add Filters


Tymmmy:
Add a new page for all transactions, no matter the PaymentPointer, no matter currency.
We already have a Transactions list set in place for PaymentPointers.
Should be grouped by date (ex. Revolut)
Add some filters on Transactions list.
For starters Filter by Outgoing/Incoming payment, date, merchant?

If needed and can be done with this story, please also add pagination to list.

raducristianpopa:
Closed by #504.

PR #504, opened 2023-07-14
feat(wallet): add general trasactions view (filters/pagination)

Fsalker:
"partially closes" #259

outdated
> 
> - ✅ gets ALL Transactions based on userId _(no **accountId** / **paymentPointerId** constraints)_
> - ✅ allows Filtering, as specified here - https://jsonapi.org/recommendations/#filtering
> - ✅ allows Pagination, using `page` and `pageSize`
> - ✅ deep filters **ought to** work out of box for **Transaction** fields
> - ❌ deep filters for **joined tables** don't currently work (yet?)¹
> 
> ### Do we really need deep filters for Joined Tables?
> 
> imo: not now
> 
> ¹We can write a **utility function** which receives **joined table names as an input** and adequately alters `filterParams`
> - eg: `prependTablenamesToFilterParams = (filterParams, tableArr) => <... newFilterParams ...>`
> 
> We could also iterate through `Transaction.relationMappings` and automatically prepend the appropriate Table Names.
> 
> But I don't think this is the case yet - `/transactions` doesn't need to return you your Accounts/PaymentPointers/User 😂, so you've got no other tables to deep-filter yet.

## Demo?

image
image
image
image
image

--- 

## Old screenshots
```
![image](https://github.com/interledger/testnet/assets/16700631/91cf5b05-c99b-451c-bf81-6b26211a9b30)
![image](https://github.com/interledger/testnet/assets/16700631/fdc2c373-b596-4d8b-a7a4-66746610138e)
![image](https://github.com/interledger/testnet/assets/16700631/b2e8dd4d-3d86-4440-af6a-5010e0b8f328)
```


packages/wallet/backend/src/transaction/validation.ts
```diff
@@ -5,3 +5,29 @@ export const transactionListRequestSchema = z.object({
...
+  id: z.string().optional(),
+  paymentId: z.string().optional(),
+  description: z.string().optional(),
+  paymentPointerId: z.string().optional(),
+  accountId: z.string().optional(),
+  assetCode: z.string().optional(),
+  value: z.bigint().optional(),
+  type: z.string().optional(),
+  status: z.string().optional()
+})
```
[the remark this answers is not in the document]
beniaminmunteanu: It's ok that you have it "redefined" in the ZOD schema.

I have a PR for rapyd zod validations that does the other way around, having types infered from ZOD. 
But for your usecase, this approach is ok. I see no trouble with it

packages/wallet/backend/src/transaction/validation.ts
```diff
@@ -5,3 +5,29 @@ export const transactionListRequestSchema = z.object({
     orderByDate: z.enum(['ASC', 'DESC']).default('DESC')
   })
 })
+
+const transactionExtendedObject = z
+  .object({
```
Tymmmy, line 10: will we need the comments?

packages/wallet/backend/src/transaction/service.ts
```diff
@@ -50,4 +59,40 @@ export class TransactionService implements ITransactionService {
...
+      }),
+      {}
+    )
+
+    const transactions = await Transaction.query()
+      .select(
+        'transactions.*',
+        'paymentPointer.url as paymentPointerUrl',
+        'paymentPointer:account.name as accountName'
+        // 'paymentPointer:account.id as accountId'
```
Tymmmy, line 84: do we need the comments?

packages/wallet/backend/src/outgoingPayment/service.ts
```diff
@@ -32,10 +34,19 @@ export class OutgoingPaymentService implements IOutgoingPaymentService {
       metadata: { description }
     })
 
+    const paymentPointer = await PaymentPointer.query()
+      .findById(paymentPointerId)
+      .select('accountId', 'active')
+
+    if (!paymentPointer || !paymentPointer.active) {
```
dragosp1011, line 41: can be simplified `if (!paymentPointer?.active)`

packages/wallet/backend/src/outgoingPayment/service.ts
```diff
@@ -32,10 +34,19 @@ export class OutgoingPaymentService implements IOutgoingPaymentService {
...
+      throw new NotFound()
+    }
+
     return Transaction.query().insert({
       paymentPointerId,
       paymentId: payment.id,
       assetCode,
+      accountId: paymentPointer?.accountId,
```
dragosp1011, line 49: as payment pointer is checked above, it will always exists here

packages/wallet/backend/src/shared/types.ts
```diff
@@ -0,0 +1,8 @@
+export type PaginationQueryParams = {
+  page: number
+  pageSize: number
+}
+
+export type ObjectWithAnyKeys = {
+  [key: string]: unknown
+}
```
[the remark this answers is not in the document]
dragosp1011: maybe use `Record<string, unknown>`?

packages/wallet/backend/src/utils/helpers.ts
```diff
@@ -1,5 +1,32 @@
...
+    (acc, oldKey) => ({
+      ...acc,
+      [oldKey === replacedKey ? newKey : oldKey]: obj[oldKey]
+    }),
+    {}
+  )
+
+export const prefixSomeObjectKeys = (
+  obj: ObjectWithAnyKeys,
+  prefixedKeys: Array<string>,
```
dragosp1011, line 20: `string[]`

packages/wallet/backend/src/utils/helpers.ts
```diff
@@ -1,5 +1,32 @@
...
+  prefix: string
+): ObjectWithAnyKeys =>
+  Object.keys(obj).reduce(
```
dragosp1011, line 23: `Object.entries` can be used instead as the key and the value is used below

packages/wallet/backend/src/utils/helpers.ts
```diff
@@ -1,5 +1,32 @@
 import { bcrypt } from 'hash-wasm'
 import { randomBytes } from 'crypto'
+import { ObjectWithAnyKeys } from '@/shared/types'
+
+export const replaceObjectKey = (
```
dragosp1011, line 5: is this used somewhere?

packages/wallet/backend/src/transaction/validation.ts
```diff
@@ -5,3 +5,22 @@ export const transactionListRequestSchema = z.object({
...
+    paymentPointerId: z.string().uuid().optional(),
+    assetCode: z.string().optional(),
+    type: z.string().optional(),
+    status: z.string().optional()
+  })
+  .strict()
+
+export const transactionListAllRequestSchema = z.object({
+  query: z.object({
+    page: z.coerce.number().int().nonnegative().default(0),
```
dragosp1011, line 21: default 1?
on first request page 1 will be returned so it make sense for default to be 1

raducristianpopa: https://vincit.github.io/objection.js/api/query-builder/other-methods.html#page

packages/wallet/backend/src/utils/helpers.ts
raducristianpopa: Not at the moment. We will keep this for the moment since we had some issues with deep relations filters.

packages/wallet/backend/src/transaction/model.ts
```diff
@@ -2,14 +2,20 @@ import { Model } from 'objection'
...
+
 export class Transaction extends BaseModel {
   static tableName = 'transactions'
 
   id!: string
   paymentId!: string
   description?: string
   paymentPointerId?: string
-  accountId?: string
+  accountId!: string
```
dragosp1011, line 18: a migration should be added to populate accountId for all existent transactions

packages/wallet/backend/src/shared/types.ts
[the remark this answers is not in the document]
raducristianpopa: I do not think there is a difference between these two.

packages/wallet/backend/src/transaction/service.ts
```diff
@@ -50,4 +60,36 @@ export class TransactionService implements ITransactionService {
...
+
+    const transactions = await Transaction.query()
+      .select(
+        'transactions.*',
+        'paymentPointer.url as paymentPointerUrl',
+        'paymentPointer:account.name as accountName'
+      )
+      .fullOuterJoinRelated('[paymentPointer.[account.user], account.user]')
+      .where(function () {
+        this.where('paymentPointer:account:user.id', userId).orWhere(
```
dragosp1011, line 84: should be simplified after adding accountId on every transaction

packages/wallet/backend/src/shared/types.ts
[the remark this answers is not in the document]
dragosp1011: there is no difference, but it is not a custom type

packages/wallet/backend/src/transaction/service.ts
raducristianpopa: I do not understand how this can simplified at the moment. 🤔 Can you elaborate?

dragosp1011: as accountId it will be always present only the second check can be kept here and remove the one with `paymentPointer:account:user.id`

packages/wallet/backend/src/transaction/controller.ts
```diff
@@ -39,4 +43,35 @@ export class TransactionController implements ITransactionController {
...
+    res: CustomResponse<Page<Transaction>>,
+    next: NextFunction
+  ) => {
+    try {
+      const userId = req.session.user.id
+      const {
+        query: { page, pageSize, filter, orderByDate }
+      } = await validate(transactionListAllRequestSchema, req)
+
+      const filterParams = (filter as Partial<Transaction>) || {}
```
dragosp1011, line 58: set the default value in the schema

dragosp1011 (dismissed):

Tymmmy (dismissed):

beniaminmunteanu (dismissed):

rico191013 (dismissed):

Tymmmy (dismissed):

raducristianpopa (approved):

PR #509, opened 2023-07-16
fix: bug that less than 1 amount is not funding or withdrawing

### LINKED ISSUE #476: Fund/WIthdraw money < 1 amounts
Smaller amounts then 1 for funds and withdraw are not going through.
We should investigate.

rico191013:
### Context

- fixes #476

### DIFF
--- a/packages/wallet/frontend/src/lib/api/account.ts
+++ b/packages/wallet/frontend/src/lib/api/account.ts
@@ -13,8 +13,7 @@ export const fundAccountSchema = z.object({
     .number({
       invalid_type_error: 'Please enter a valid amount'
     })
-    .min(0, { message: 'Please enter an amount' })
-    .positive()
+    .nonnegative({ message: 'Please enter an amount' })
 })
 
 export const withdrawFundsSchema = fundAccountSchema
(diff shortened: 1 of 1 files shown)


packages/wallet/frontend/src/lib/api/account.ts
```diff
@@ -13,7 +13,7 @@ export const fundAccountSchema = z.object({
     .number({
       invalid_type_error: 'Please enter a valid amount'
     })
-    .min(1, { message: 'Please enter an amount' })
+    .min(0, { message: 'Please enter an amount' })
     .positive()
```
raducristianpopa, line 17: [Zod numbers](https://zod.dev/?id=numbers).
```suggestion
    .nonnegative({ message: 'Please enter an amount' })
```

raducristianpopa (dismissed):

packages/wallet/frontend/src/lib/api/account.ts
```diff
@@ -13,8 +13,7 @@ export const fundAccountSchema = z.object({
     .number({
       invalid_type_error: 'Please enter a valid amount'
     })
-    .min(1, { message: 'Please enter an amount' })
-    .positive()
+    .nonnegative({ message: 'Please enter an amount' })
```
raducristianpopa, line 16: My fault. It should be `positive` instead of `nonnegative`. `nonnegative` allows number that are >= 0.
```suggestion
    .positive({ message: 'Please enter an amount' })
```

beniaminmunteanu (approved):

raducristianpopa (approved):

PR #527, opened 2023-07-20
feat(wallet): allow the user to set the expiry of a payment

### LINKED ISSUE #475: Add an input to determine expiry date for payments
From feedback, let the user set the expiry date of the payment.
Only for Request payment.

raducristianpopa:
### Context

- closes #475 

### Changes
- allow the user to specify the duration after which the payment should expire


raducristianpopa (commented):
The PR is still a work in progress and requires some final touches, but I welcome any initial feedback on it. Currently, I'm awaiting some information regarding the expiry of an incoming payment.

During my testing of this new functionality, I may have come across a potential bug. I will investigate further and create an issue to address it.

packages/wallet/backend/src/incomingPayment/service.ts
```diff
@@ -129,4 +139,16 @@ export class IncomingPaymentService implements IIncomingPaymentService {
 
     return `${existingPaymentPointer.url}/incoming-payments/${response.paymentId}`
   }
+
+  private generateExpiryObject(expiry: number, unit: string): Duration {
+    switch (unit) {
+      case 's': {
```
rico191013, line 145: this function also supports 'm' and 'h' as well?

raducristianpopa: Yes, this will also support `m(inutes)`, `h(ours)` and `d(ays)`.

packages/wallet/backend/migrations/20230728101910_update_transactions_columns.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+exports.up = function (knex) {
+  return knex.schema.alterTable('transactions', async (table) => {
+    table.timestamp('expiresAt')
+    await knex.raw(
+      [
+        `ALTER TABLE "transactions" DROP CONSTRAINT IF EXISTS "transactions_status_check";`,
+        `ALTER TABLE "transactions" ADD CONSTRAINT "transactions_status_check" CHECK ("status" = ANY (ARRAY['PENDING'::text, 'COMPLETED'::text, 'REJECTED'::text, 'EXPIRED'::text]))`
+      ].join('\n')
+    )
```
raducristianpopa, line 13: `knex` does not have support to alter `enums` since they are only some check constraints.

packages/wallet/backend/src/transaction/service.ts
```diff
@@ -90,4 +93,31 @@ export class TransactionService implements ITransactionService {
 
     return transactions
   }
+
+  async processPendingIncomingPayments(): Promise<string | undefined> {
+    return this.deps.knex.transaction(async (trx) => {
+      const now = new Date(Date.now())
+      const [transaction] = await Transaction.query(trx)
```
dragosp1011, line 100: can all expired transactions be updated at the same time?

packages/wallet/backend/src/app.ts
```diff
@@ -269,4 +273,24 @@ export class App {
...
     return app
   }
+
+  private async processPendingTransactions() {
+    const transactionService = await this.container.resolve(
+      'transactionService'
+    )
+    return transactionService
+      .processPendingIncomingPayments()
+      .catch(() => true)
```
dragosp1011, line 283: in case of an error I think it should be a delay, so maybe here return false

packages/wallet/backend/src/app.ts
```diff
@@ -269,4 +273,24 @@ export class App {
...
+      .then((trx) => {
+        if (trx) {
+          process.nextTick(() => this.processPendingTransactions())
+        } else {
+          setTimeout(() => this.processPendingTransactions(), 5000).unref()
```
dragosp1011, line 288: I think the interval is too small, it is not as important to have the status updated as it does not influence at anything so it easily can be 1 or even 5 mins

packages/wallet/backend/src/incomingPayment/validation.ts
```diff
@@ -1,11 +1,24 @@
...
-    paymentPointerId: z.string().uuid(),
-    amount: z.number().positive(),
-    description: z.string().optional()
-  })
+  body: z
+    .object({
+      paymentPointerId: z.string().uuid(),
+      amount: z.number().positive(),
+      description: z.string().optional(),
+      expiry: z.coerce.number().positive().int().optional(),
```
dragosp1011, line 9: I suggest to have an optional object for expiration, and the value and unit to be required inside the object

packages/wallet/backend/src/app.ts
```diff
@@ -87,6 +87,10 @@ export class App {
     })
     Model.knex(knex)
 
+    if (env.NODE_ENV !== 'test') {
+      this.processResources()
```
dragosp1011, line 91: how about using a cron job lib for scheduling the worker?

packages/wallet/backend/src/incomingPayment/service.ts
```diff
@@ -132,4 +143,23 @@ export class IncomingPaymentService implements IIncomingPaymentService {
 
     return `${existingPaymentPointer.url}/incoming-payments/${response.paymentId}`
   }
+
+  private generateExpiryObject(expiry: number, unit: string): Duration {
+    switch (unit) {
```
dragosp1011, line 148: an object mapping can be used
```
const mapping = {
s: 'seconds',
...
}
return mapping[unit] ? { mapping[unit]: expiry } : { days: 30 }
```

packages/wallet/backend/src/transaction/service.ts
raducristianpopa: I don't really like the idea of mass-updating the transactions. 🤔

packages/wallet/backend/src/incomingPayment/validation.ts
```diff
@@ -1,11 +1,32 @@
...
+        .object({
+          value: z.coerce.number().positive().int(),
+          unit: z.enum(['s', 'm', 'h', 'd'])
+        })
+        .optional()
+    })
+    .superRefine(({ expiration }, ctx) => {
+      if (
+        expiration &&
+        ((expiration.value && !expiration.unit) ||
```
dragosp1011, line 19: these checks are not required after moving the expiration into an object

packages/wallet/frontend/src/lib/api/transfers.ts
```diff
@@ -27,18 +27,58 @@ export const acceptQuoteSchema = z.object({
...
+      .object({
+        value: z.string().uuid(),
+        label: z.string().min(1)
+      })
+      .nullable(),
+    amount: z.coerce.number({
+      invalid_type_error: 'Please enter a valid amount'
+    }),
+    description: z.string(),
+    expiry: z.coerce
```
dragosp1011, line 46: change this to the same format as on backend

dragosp1011 (approved):

PR #539, opened 2023-07-26
feat(wallet-frontend): Display the exchange currency when working with cross currency transa…

### LINKED ISSUE #536: Display the converted amount for cross-currency transactions
In cross-currency transactions, we should show the user the converted amount. For example, in a scenario where the sender wants to send 10 EUR to USD, we should display a field with the converted amount to USD. This will only serve as an information for a better UX. The converted amount should not be sent to the backend.

Tymmmy:
…ctions

### Context

- fixes #536 

### Changes
If the Transaction being set up is between cross currency payments, then a short info will appear for the user bellow amount, to inform the user about the exchange rat being applied in the transaction.

### DIFF
--- a/packages/wallet/frontend/src/lib/api/paymentPointer.ts
+++ b/packages/wallet/frontend/src/lib/api/paymentPointer.ts
@@ -74,6 +74,10 @@ type BasePaymentPointerArgs = {
   paymentPointerId: string
 }
 
+type PaymentPointeAssetCode = {
+  assetCode: string
+}
+
 type GetPaymentPointerArgs = { accountId: string; paymentPointerId: string }
 type GetPaymentPointerResult = SuccessResponse<PaymentPointer>
 type GetPaymentPointerResponse = GetPaymentPointerResult | ErrorResponse
@@ -112,6 +116,9 @@ type GenerateKeyResponse = GenerateKeyResult | ErrorResponse
 type RevokeKeyArgs = BasePaymentPointerArgs
 type RevokeKeyResponse = SuccessResponse | ErrorResponse
 
+type AssetCodeResult = SuccessResponse<PaymentPointeAssetCode>
+type AssetCodeResponse = AssetCodeResult | ErrorResponse
+
 interface PaymentPointerService {
   get: (
     args: GetPaymentPointerArgs,
@@ -135,6 +142,7 @@ interface PaymentPointerService {
   ) => Promise<ListTransactionsResponse>
   generateKey: (args: GenerateKeyArgs) => Promise<GenerateKeyResponse>
   revokeKey: (args: RevokeKeyArgs) => Promise<RevokeKeyResponse>
+  assetCode: (url: string, cookies?: string) => Promise<AssetCodeResponse>
 }
 
 const createPaymentPointerService = (): PaymentPointerService => ({
@@ -277,6 +285,21 @@ const createPaymentPointerService = (): PaymentPointerService => ({
         error,
         'We were not able to revoke the key. Please try again.'
       )
+    }
+  },
+
+  async assetCode(url, cookies) {
+    try {
+      const response = await httpClient
+        .get(`external-payment-pointers?url=${url}`, {
+          headers: {
+            ...(cookies ? { Cookie: cookies } : {})
+          }
+        })
+        .json<AssetCodeResult>()
+      return response
+    } catch (error) {
+      return getError(error, 'Error fetching payment pointer url asset code.')
     }
   }
 })
--- a/packages/wallet/frontend/src/lib/api/transfers.ts
+++ b/packages/wallet/frontend/src/lib/api/transfers.ts
@@ -43,6 +43,7 @@ export const requestSchema = z.object({
 type PaymentDetails = {
   description?: string
   value: number
+  assetCode: string
 }
 
 type AmountProps = {
--- a/packages/wallet/frontend/src/pages/transfer/send.tsx
+++ b/packages/wallet/frontend/src/pages/transfer/send.tsx
@@ -12,9 +12,9 @@ import { GetServerSideProps, InferGetServerSidePropsType } from 'next'
 import { accountService } from '@/lib/api/account'
 import { sendSchema, transfersService } from '@/lib/api/transfers'
 import { SuccessDialog } from '@/components/dialogs/SuccessDialog'
-import { formatAmount, getObjectKeys } from '@/utils/helpers'
+import { formatAmount, getCurrencySymbol, getObjectKeys } from '@/utils/helpers'
 import { useDialog } from '@/lib/hooks/useDialog'
-import { useEffect, useState } from 'react'
+import { ChangeEvent, useEffect, useState } from 'react'
 import { paymentPointerService } from '@/lib/api/paymentPointer'
 import { ErrorDialog } from '@/components/dialogs/ErrorDialog'
 import { Controller } from 'react-hook-form'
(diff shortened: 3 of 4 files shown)


packages/wallet/frontend/src/lib/api/paymentPointer.ts
```diff
@@ -74,6 +74,10 @@ type BasePaymentPointerArgs = {
   paymentPointerId: string
 }
 
+type PaymentPointeAssetCode = {
+  assetCode: string
+}
```
raducristianpopa, line 79:
```suggestion
type PaymentPointerAssetCode = {
  assetCode: string
}
```

I would also suggest renaming this to `PaymentPointerOP` (Payment Pointer Open Payments) and ensuring that all fields are typed, not just the asset code. Additionally, it's worth noting that the receiver account can have a different asset scale than the sender.

packages/wallet/frontend/src/lib/api/paymentPointer.ts
```diff
@@ -135,6 +142,7 @@ interface PaymentPointerService {
   ) => Promise<ListTransactionsResponse>
   generateKey: (args: GenerateKeyArgs) => Promise<GenerateKeyResponse>
   revokeKey: (args: RevokeKeyArgs) => Promise<RevokeKeyResponse>
+  assetCode: (url: string, cookies?: string) => Promise<AssetCodeResponse>
```
raducristianpopa, line 145: Can not think of a proper name at the moment, but this should be changed.
```suggestion
  getExternal: (url: string, cookies?: string) => Promise<AssetCodeResponse>
```

packages/wallet/frontend/src/pages/transfer/send.tsx
```diff
@@ -26,13 +26,19 @@ import {
...
 type SendProps = InferGetServerSidePropsType<typeof getServerSideProps>
 
 const SendPage: NextPageWithLayout<SendProps> = ({ accounts }) => {
   const [openDialog, closeDialog] = useDialog()
   const [paymentPointers, setPaymentPointers] = useState<SelectOption[]>([])
   const [balance, setBalance] = useState('')
+  const [currentExchangeRates, setCurrentExchangeRates] = useState<Rates>()
+  const [selectedAssetCode, setSelectedAssetCode] = useState('')
+  const [receiverAssetCode, setReceiverAssetCode] = useState('')
+  const [exchangeCurrencyText, setExchangeCurrencyText] = useState('')
```
raducristianpopa, line 40: I think we should type this as `string | null` instead of an empty string.

packages/wallet/frontend/src/pages/transfer/send.tsx
```diff
@@ -280,6 +344,7 @@ const SendPage: NextPageWithLayout<SendProps> = ({ accounts }) => {
                 />
               }
             />
+            <p className="ml-2 text-sm text-green">{exchangeCurrencyText}</p>
```
raducristianpopa, line 347:
```suggestion
            {exchangeCurrencyText ? <p className="ml-2 text-sm text-green">{exchangeCurrencyText}</p> : null}
```

packages/wallet/frontend/src/pages/transfer/send.tsx
```diff
@@ -108,16 +150,37 @@ const SendPage: NextPageWithLayout<SendProps> = ({ accounts }) => {
...
+      setReceiverAssetCode(paymentPointerAssetCodeResponse.data.assetCode)
     }
 
     if (isToggleDisabled) setIsToggleDisabled(false)
 
     sendForm.setValue('receiver', url)
   }
 
+  const onAmountChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setConvertAmount(Number(event.currentTarget.value))
```
raducristianpopa, line 181: We should validate `event.currentTarge.value`.

packages/wallet/frontend/src/pages/transfer/send.tsx
raducristianpopa, line 40: I don't know what to say about having another state action for displaying some text. I think we can do this with conditional rendering.

packages/wallet/frontend/src/lib/api/paymentPointer.ts
raducristianpopa: Talking about this, I believe we might have a bug when a user adds an incoming payment URL as the receiver. If the incoming payment URL has a different asset, the Amount field gets populated with the amount that the incoming payment was initially created with. For example:
- Bob creates an incoming payment for 10 EUR
- Alice wants to send money to Bob from a USD account into that incoming payment
- When Alice adds the incoming payment URL as the receiver, we fetch the incoming payment details
- We populate the Amount field with the amount that was specified in the incoming payment

This will result in Alice sending Bob 10 USD, which is actually less than 10 EUR.

packages/wallet/frontend/src/lib/api/paymentPointer.ts
```diff
@@ -74,6 +74,14 @@ type BasePaymentPointerArgs = {
...
 }
 
+type PaymentPointeOP = {
+  id: string
+  assetCode: string
+  assetScale: number
+  publicName: string
+  authServer: string
+}
+
```
raducristianpopa, line 84: Add an `AssetOP` in `lib/api/asset.ts` and import it from there:
```suggestion
type PaymentPointerOP = AssetOP & {
  id: string
  publicName: string
  authServer: string
}

```

packages/wallet/frontend/src/lib/api/transfers.ts
```diff
@@ -43,6 +43,7 @@ export const requestSchema = z.object({
 type PaymentDetails = {
   description?: string
   value: number
+  assetCode: string
 }
```
raducristianpopa, line 47:
```suggestion
type PaymentDetails = AssetOP & {
  description?: string
  value: number
}
```

packages/wallet/frontend/src/pages/transfer/send.tsx
```diff
@@ -26,16 +26,48 @@ import {
...
+          receiverAssetCode
+        )}. Exchange Rate: ${getCurrencySymbol(
+          selectedAsset.assetCode
+        )}${convertAmount} = ${getCurrencySymbol(receiverAssetCode)}${(
+          convertAmount * currentExchangeRates.rates[receiverAssetCode]
+        ).toFixed(selectedAsset.assetScale)}`}</p>
+      )
+    } else return null
+  })
+  ExchangeRate.displayName = 'ExchangeRate'
```
raducristianpopa, line 69: You can move this outside the `SendPage` component and pass in the props that it needs for displaying this - `receiverAssetCode`, `convertAmount`, etc...

packages/wallet/frontend/src/pages/transfer/send.tsx
```diff
@@ -26,16 +26,48 @@ import {
 } from '@/utils/constants'
 import { useOnboardingContext } from '@/lib/context/onboarding'
 import { QuoteDialog } from '@/components/dialogs/QuoteDialog'
+import { assetService, Rates } from '@/lib/api/asset'
 
 type SendProps = InferGetServerSidePropsType<typeof getServerSideProps>
 
+type Asset = { assetCode: string; assetScale: number }
```
raducristianpopa, line 33: Ditto above, import the type from `lib/api/asset.ts`.
```suggestion
```

raducristianpopa:
Postponed at the moment.

Tymmmy:
Many conflicts, decided to start again with clean branch from main.

PR #600, opened 2023-08-09
feat: added daily rates api

### LINKED ISSUE #535: Call Rapyd Daily Rates API
Call in BE Rapyd daily rates API:
https://docs.rapyd.net/build-with-rapyd/reference/daily-rates

It will be called in the cross currency story.

rico191013:
### Context

- fixes #535


packages/wallet/backend/src/rapyd/controller.ts
```diff
@@ -182,4 +187,25 @@ export class RapydController implements IRapydController {
...
+  ) => {
+    try {
+      const {
+        query: { base }
+      } = await validate(ratesSchema, req)
+      const rates = await this.deps.rapydService.getRates(base)
+
+      res.status(200).json({
+        success: true,
+        message: 'Profile updated succesfully',
```
Tymmmy, line 204: Is this message correct? I don't think we are updating profile here....

beniaminmunteanu (approved):

beniaminmunteanu:
maybe we can find a better place for the daily rates outside of rapyd, since it is no longer using it. But it is ok for now.

PR #604, opened 2023-08-10
chore: revert prettier config

dragosp1011:
Reverts interledger/testnet#579
This was causing issues on deploy. Needs more investigation.


dragosp1011:
the issue was not reproduced, most likely the change triggered rebuild of some cached layers that caused the trouble and it is not the source of the problem.

PR #636, opened 2023-08-21
feat(wallet-backend): Integrate Daily rates in cross currency transfers

beniaminmunteanu:
This PR makes use of the daily rates API and provides the rates to Rafiki for quote creation.


packages/wallet/backend/src/quote/service.ts
```diff
@@ -168,10 +168,22 @@ export class QuoteService implements IQuoteService {
...
-    ]
-    return BigInt((Number(params.amount) * conversionRate).toFixed())
+  private async convert(params: ConvertParams): Promise<bigint> {
+    const conversionRate = await this.deps.rafikiService.getRates(params.from)
+    return BigInt(
+      (Number(params.amount) * conversionRate.rates[params.to]).toFixed()
+    )
   }
+
+  // private async convert(params: ConvertParams): Promise<bigint> {
```
Tymmmy, line 178: do we need the comments?

beniaminmunteanu: leftover. Thx

packages/wallet/backend/src/createContainer.ts
```diff
@@ -265,10 +265,12 @@ export const createContainer = (config: Env): Container<Bindings> => {
     const logger = await container.resolve('logger')
     const rafikiClient = await container.resolve('rafikiClient')
     const transactionService = await container.resolve('transactionService')
+    const rapydService = await container.resolve('rapydService')
```
raducristianpopa, line 268: Why are we using `rapydService` here?

beniaminmunteanu: because that is where the daily rates retrieval is done.
Rico had it inside the Rapyd module in his PR, and since we didn't have a better place for it yet, I was ok with it.
We could extract it somewhere else in the future I guess.

beniaminmunteanu: The two options would be a separate module for rates, or a ratesService inside QuoteService.
I'd go for the separate module, since rates are also going to be used in Exchange, not just in Quotes.
@rico191013 ?

beniaminmunteanu: moved it into a separate module, please have a look

packages/wallet/backend/src/app.ts
```diff
@@ -232,7 +237,7 @@ export class App {
     router.post('/wallet', isAuth, rapydController.createWallet)
     router.post('/updateProfile', isAuth, rapydController.updateProfile)
     router.post('/verify', isAuth, rapydController.verifyIdentity)
-    router.get('/rates/daily', rapydController.getRates)
+    router.get('/rates/daily', ratesController.getRates)
```
raducristianpopa, line 240: What's the particular use case for this endpoint ?

beniaminmunteanu: Is the frontend not using it? I thought we will have the frontend try to do the conversion beforehand, when the "receive" is toggled?

packages/wallet/backend/src/rafiki/service.ts
```diff
@@ -500,18 +499,7 @@ export class RafikiService implements IRafikiService {
...
-      },
-      EUR: {
-        USD: 1.10019
-      }
-    }
-    return {
-      base,
-      rates: exchangeRates[base] ?? {}
-    }
+  public async getRates(base: string): Promise<RatesResponse> {
```
rico191013, line 502: Do we need this function in rafiki.service? maybe the caller can directly call from ratesService.

beniaminmunteanu: indeed.

packages/wallet/backend/src/app.ts
beniaminmunteanu: got it out of the way, since frontend is using /rates -> the endpoint created for rafiki

raducristianpopa (commented):
LGTM. I think we should add the caching mechanism in this PR as well.

packages/wallet/backend/src/rates/service.ts
```diff
@@ -0,0 +1,24 @@
+import axios from 'axios'
+
+export type RatesResponse = {
+  base: string
+  rates: Record<string, number>
+}
+
+export interface IRatesService {
+  getRates: (userId: string) => Promise<RatesResponse>
```
dragosp1011, line 9: userId?

Tymmmy: @beniaminmunteanu  looks like it should be base, not userId

beniaminmunteanu:
> LGTM. I think we should add the caching mechanism in this PR as well.

I'd rather have this go in first...

Tymmmy (dismissed):

raducristianpopa:
> > LGTM. I think we should add the caching mechanism in this PR as well.
> 
> I'd rather have this go in first...

Let's make sure we don't overwhelm this free API right from the start, especially when we're testing payments repeatedly.

beniaminmunteanu:
> > > LGTM. I think we should add the caching mechanism in this PR as well.
> > 
> > 
> > I'd rather have this go in first...
> 
> Let's make sure we don't overwhelm this free API right from the start, especially when we're testing payments repeatedly.

This was a valid point.

I Added an in-memory cache package for now and cached responses for 12 hours, in order to take some pressure out of the free api while testing.

There will be a future PR with more effort into caching

dragosp1011 (approved):

Tymmmy (approved):

PR #639, opened 2023-08-22
feat: Automatic balance update

### LINKED ISSUE #301: Automatic Balance Update & Notification on receiving money
Implement websocket? or something else to be used in automatic Balance update.

How it is working now?
 - User A sends money to user B
 - User B will not realize that he got money, only if they go to transactions, or reload page and see that balance has more money.

How should it work?
- User A send money to User B
- User B receives a notification (for example a small popup bottom right corner or animation) that he received some money.
- Balance of User B is automatically updated, without reload of page
- User B has been notified that he received money, so he can go to transactions and see details.

(Also use some animation/notification when receiving money?)

### LINKED ISSUE #690: Implement FE part on websocket notifications
This story can be used to implement FE part of the story:
https://github.com/interledger/testnet/issues/301

Add FE notifications, popus when receiving money.

rico191013:
### Context

- fixes #301
- fixes #690


packages/wallet/backend/src/socket/service.ts
```diff
@@ -0,0 +1,58 @@
...
+          'http://localhost:4003',
+          `https://${this.deps.env.RAFIKI_MONEY_FRONTEND_HOST}`
+        ],
+        credentials: true
+      }
+    })
+    this.deps.logger.info(`Socket Server is started...`)
+
+    this.io.on('connection', (socket) => {
+      const email = socket.handshake.query.email // Access the user's email
```
beniaminmunteanu, line 35: are you including auth credentials in the socket handshake? or how will the email be provided here?

packages/wallet/frontend/src/pages/index.tsx
```diff
@@ -16,10 +16,42 @@ import { userService } from '@/lib/api/user'
...
 type HomeProps = InferGetServerSidePropsType<typeof getServerSideProps>
 
 const HomePage: NextPageWithLayout<HomeProps> = ({ accounts, user }) => {
+  let socket: Socket
+
+  useEffect(() => {
+    // Access cookies
+
+    // Connect to the Socket.IO server
+    socket = io(process.env.NEXT_PUBLIC_BACKEND_URL ?? '', {
```
beniaminmunteanu, line 30: check the linter suggestion here.

packages/wallet/backend/src/socket/service.ts
```diff
@@ -0,0 +1,58 @@
...
+      this.sockets.set(email, socket)
```
beniaminmunteanu, line 36: Socket.io provides a built-in concept of Rooms. Instead of directly storing sockets in your own data structure, we could use rooms to manage connections. When a user connects, we can add their socket to a room named after their email. This way, we don't need to explicitly manage the sockets ourselves.

```
io.on('connection', (socket) => {
    const email = socket.handshake.query.email
    socket.join(userEmail);
});
```

packages/wallet/backend/src/socket/service.ts
```diff
@@ -0,0 +1,58 @@
...
+      socket.on('disconnect', () => {
+        this.sockets.delete(email)
+        this.deps.logger.info(`socket client disconnected...`)
+      })
+    })
+  }
+
+  emitBalanceUpdateByEmail(email: string) {
+    if (this.io) {
+      const socket = this.sockets.get(email)
```
beniaminmunteanu, line 49: and here, it could be:

`io.to(email).emit('message', 'Hello there!');
`
if approach described above is used

packages/wallet/backend/src/socket/service.ts
dragosp1011: rooms is the way to go, and I suggest to use user id instead of email

packages/wallet/frontend/src/pages/index.tsx
```diff
@@ -16,10 +16,42 @@ import { userService } from '@/lib/api/user'
...
+      query: {
+        email: user.email
```
dragosp1011, line 32: instead of passing the email, the authorization token should be set in request and the user info will be extracted from there on authorization

packages/wallet/frontend/src/utils/crypto.ts
```diff
@@ -0,0 +1,22 @@
+import cryptolib from 'crypto'
+
+const secretKeyHex =
+  '00112233445566778899aabbccddeeff00112233445566778899aabbccddeeff'
+const iv = 'rafiki-testnet23'
+
+const encrypt = (text: string): string => {
```
raducristianpopa, line 7: I'm not sure about this approach. Since the `encrypt` function is used on the client-side it will also have `secretKeyhex` and `iv` in the final client bundle.

packages/wallet/backend/src/socket/service.ts
```diff
@@ -0,0 +1,60 @@
...
+          'http://localhost:4003',
+          `https://${this.deps.env.RAFIKI_MONEY_FRONTEND_HOST}`
+        ],
+        credentials: true
+      }
+    })
+    this.deps.logger.info(`Socket Server is started...`)
+
+    this.io.on('connection', async (socket) => {
+      const token = socket.handshake.query.token?.toString().split('=')[1]
```
dragosp1011, line 37: can the token be sent in headers or cookies?

rico191013: hmm, in my understanding handshake is used for this purpose in socket communication. how do you think @beniaminmunteanu ?

beniaminmunteanu: I also think it should be in the auth headers, and checked on the handshake handle. 
 The client already has the auth token stored when the user logs in. the same token should be used when creating the websocket connection, and as such checked at handshake time.

packages/wallet/frontend/src/pages/index.tsx
```diff
@@ -16,10 +16,42 @@ import { userService } from '@/lib/api/user'
...
+  // const [socket, setSocket] = useState<Socket | null>(null)
+
+  useEffect(() => {
+    let socket: Socket | null = null
+    // Connect to the Socket.IO server
+    socket = io(process.env.NEXT_PUBLIC_BACKEND_URL ?? '', {
+      query: {
+        token
+      }
+    })
```
casalete, line 33:
```suggestion
     socket = io(process.env.NEXT_PUBLIC_BACKEND_URL ?? '', {
       auth: {token} 
  });
```

casalete: use auth for it.

packages/wallet/backend/src/socket/service.ts
```diff
@@ -0,0 +1,63 @@
...
```
casalete, line 37:
```suggestion
      const token = socket.handshake.auth.token?.toString().split('=')[1]
```

packages/wallet/frontend/src/pages/index.tsx
beniaminmunteanu, line 33:
```suggestion
    socket = io(process.env.NEXT_PUBLIC_BACKEND_URL ?? '', {auth: {token}})
```

beniaminmunteanu: use auth for it, it's not a query.

packages/wallet/backend/src/socket/service.ts
beniaminmunteanu, line 37:
```suggestion
      const token = socket.handshake.auth.token?.toString().split('=')[1]
```

beniaminmunteanu: by using the auth, it will be sent as a header @dragosp1011 on the initial HTTP upgrade request

beniaminmunteanu:
https://socket.io/docs/v4/middlewares/#sending-credentials

Have a look at this piece of docs. @dragosp1011  i've seen the extraHeaders implementation before. But on newer socket.io versions, this is the recommended way. Since the extra-headers somehow did not work for all browsers.

packages/wallet/backend/src/socket/service.ts
dragosp1011: agree with this approach, looks like it is the recommended way

packages/wallet/frontend/src/pages/index.tsx
```diff
@@ -16,10 +16,42 @@ import { userService } from '@/lib/api/user'
...
+    let socket: Socket | null = null
+    // Connect to the Socket.IO server
+    socket = io(process.env.NEXT_PUBLIC_BACKEND_URL ?? '', {
+      auth: {
+        token
+      }
+    })
+
+    // Event listeners
+    socket?.on('connect', () => {
```
Tymmmy, line 36: should we leave here the console logs ?

beniaminmunteanu:
I think some handles for connection errors and reconnection logic should be added as well. Check the socketIo docs on both client and server events, there are other events such as :
**Server-Side**:

- error: Fired when an error occurs, either during connection, data exchange, or any other operation.

**Client-Side Events**:

- disconnect: Fired when the client disconnects from the server, either intentionally or due to an error.

- connect_error: Fired when an error occurs during the connection process.

- connect_timeout: Fired when a connection attempt times out.

- reconnect: Fired when the client successfully reconnects after a disconnection.

- reconnecting: Fired when the client is in the process of attempting to reconnect.

- reconnect_error: Fired when an error occurs during the reconnection process.

- reconnect_failed: Fired when the client exhausts its reconnection attempts.


Also, I would find a way to centralize our available websocket messages, which is now only "ACCOUNTS_UPDATE", so we can use them from an ENUM and be able to document the WS communication.

Make sure to release the connection when a user is logged out or kicked out due to unauthorized requests or token expiry

packages/wallet/backend/src/socket/service.ts
```diff
@@ -0,0 +1,63 @@
...
+      })
+    })
+  }
+
+  async emitAccountsUpdateById(userId?: string) {
+    if (!userId) return
+
+    const accounts = await this.deps.accountService.getAccounts(userId, false)
+
+    console.log('accounts', accounts)
```
Tymmmy, line 59: should we leave the console logs in?

Tymmmy (changes requested):
Is it possible to have a notification when someon receives money? And not just the balance update? I would like to specify that YOU RECEIVED MONEY. If I use currenct Account updated event, that is also triggers for the one sending the money, as the account updates. Would that be possible to differenciate between events?

packages/wallet/backend/src/socket/service.ts
```diff
@@ -0,0 +1,62 @@
...
+          'http://localhost:4003',
+          `https://${this.deps.env.RAFIKI_MONEY_FRONTEND_HOST}`
+        ],
+        credentials: true
+      }
+    })
+    this.deps.logger.info(`Socket Server is started...`)
+
+    this.io.on('connection', async (socket) => {
+      const cookie = socket.request.headers['cookie']
```
dragosp1011, line 37: as the token is sent in cookies it should be possible to use the express middleware
https://socket.io/docs/v4/middlewares/#:~:text=Compatibility%20with%20Express%20middleware%E2%80%8B,really%20compatible%20with%20Express%20middlewares.

rico191013:
> Is it possible to have a notification when someon receives money? And not just the balance update? I would like to specify that YOU RECEIVED MONEY. If I use current Account updated event, that is also triggers for the one sending the money, as the account updates. Would that be possible to differenciate between events?

I added MONEY_RECEIVED, and MONEY_SENT event. Would this be fine?

rico191013:
> I think some handles for connection errors and reconnection logic should be added as well. Check the socketIo docs on both client and server events, there are other events such as : **Server-Side**:
> 
> * error: Fired when an error occurs, either during connection, data exchange, or any other operation.
> 
> **Client-Side Events**:
> 
> * disconnect: Fired when the client disconnects from the server, either intentionally or due to an error.
> * connect_error: Fired when an error occurs during the connection process.
> * connect_timeout: Fired when a connection attempt times out.
> * reconnect: Fired when the client successfully reconnects after a disconnection.
> * reconnecting: Fired when the client is in the process of attempting to reconnect.
> * reconnect_error: Fired when an error occurs during the reconnection process.
> * reconnect_failed: Fired when the client exhausts its reconnection attempts.
> 
> Also, I would find a way to centralize our available websocket messages, which is now only "ACCOUNTS_UPDATE", so we can use them from an ENUM and be able to document the WS communication.
> 
> Make sure to release the connection when a user is logged out or kicked out due to unauthorized requests or token expiry

- For error cases, is there anything to do except logging that error to the console? @beniaminmunteanu

packages/wallet/backend/src/socket/service.ts
```diff
@@ -0,0 +1,70 @@
...
+        this.deps.logger.error('Error in socket communication', e)
+      })
+    })
+  }
+
+  async emitMoneyReceivedByUserId(userId?: string) {
+    if (!userId) return
+
+    const accounts = await this.deps.accountService.getAccounts(userId, false)
+    this.io?.to(userId).emit(MessageType.MONEY_RECEIVED, accounts)
```
dragosp1011, line 61: I don't think the events should send all the accounts
The account id + updated balance should be enough
there are a few other options: send the account id (frontend will fetch it) or send the incoming payment id (frontend can fetch the incoming payment and can show with payment received money)

same applies to sent money event

docker/dev/docker-compose.yml
```diff
@@ -111,7 +111,7 @@ services:
       CONNECTOR_PORT: 3002
       OPEN_PAYMENTS_PORT: 80
       DATABASE_URL: postgresql://rafiki_backend:rafiki_backend@postgres/rafiki_backend
-      USE_TIGERBEETLE: true
+      USE_TIGERBEETLE: false
```
dragosp1011, line 114: why?

dragosp1011 (approved):

docker/dev/docker-compose.yml
rico191013: well, tigerbeetle was not working on my linux since some weeks ago.

dragosp1011: it can be changed on your machine, but should remain to true in the repo

packages/wallet/frontend/src/pages/_app.tsx
```diff
@@ -13,6 +18,41 @@ const titilium = Titillium_Web({
...
+      withCredentials: true
+    })
+
+    // Event listeners
+    socket?.on('connect', () => {
+      // console.log('Connected to server')
+    })
+
+    socket?.on('MONEY_RECEIVED', (data) => {
+      toast({
```
beniaminmunteanu, line 36: what happens if you are on a page in which you see the balance somewhere?
That should be extracted as some sort of global state, and this should update it as well.

packages/wallet/frontend/src/pages/_app.tsx
```diff
@@ -13,6 +18,41 @@ const titilium = Titillium_Web({
 
 export default function App({ Component, pageProps }: AppPropsWithLayout) {
   const getLayout = Component.getLayout ?? ((page) => page)
+  const { toast } = useToast()
+
+  useEffect(() => {
+    let socket: Socket | null = null
+    // Connect to the Socket.IO server
+    socket = io(process.env.NEXT_PUBLIC_BACKEND_URL ?? '', {
```
beniaminmunteanu, line 30: provide config for automatically reconnecting, and maybe fallback option to long-polling

packages/wallet/frontend/src/pages/_app.tsx
Tymmmy: Just updated with the code to automatically update balance

beniaminmunteanu:
> > I think some handles for connection errors and reconnection logic should be added as well. Check the socketIo docs on both client and server events, there are other events such as : **Server-Side**:
> > 
> > * error: Fired when an error occurs, either during connection, data exchange, or any other operation.
> > 
> > **Client-Side Events**:
> > 
> > * disconnect: Fired when the client disconnects from the server, either intentionally or due to an error.
> > * connect_error: Fired when an error occurs during the connection process.
> > * connect_timeout: Fired when a connection attempt times out.
> > * reconnect: Fired when the client successfully reconnects after a disconnection.
> > * reconnecting: Fired when the client is in the process of attempting to reconnect.
> > * reconnect_error: Fired when an error occurs during the reconnection process.
> > * reconnect_failed: Fired when the client exhausts its reconnection attempts.
> > 
> > Also, I would find a way to centralize our available websocket messages, which is now only "ACCOUNTS_UPDATE", so we can use them from an ENUM and be able to document the WS communication.
> > Make sure to release the connection when a user is logged out or kicked out due to unauthorized requests or token expiry
> 
> * For error cases, is there anything to do except logging that error to the console? @beniaminmunteanu


@dragosp1011 
@raducristianpopa 

Do you think we should also log the user out in certain cases?

For example, if recoonect_failed happens on the client (when a given number of reconnect attempts fail) should we clear it's token from the browser somehow and redirect the user to login?

Should we display an overlay over the frontend application while the websocket is "reconnecting" ?

Also, from the server's point of view.
Upon the 'error' event fired, should we log the user out, so that he gets "unauthenticated" at his next request, and gets redirected to the login page?

If we say no to any of these, then I guess the console logs are useless and we can not react on the socket.io events at all

Tymmmy (dismissed):

Tymmmy:
@beniaminmunteanu I added the connect_error handling for now, with a reconnection attempt.
Beacuse this is used for now to notify users about money coming in, and automatic balance update, I would not force a logout of the user just for this. They will not even notice the missing notification, but they will definetly notice being loged out :D. That's my opinion :D . So I think a socket.connect triggering again should be enough? (for now)

dragosp1011:
agree with Timea, the events are not critical and logging the user out does not look necessary. The user should be logged out on next backend request anyway if something is wrong with the token.
Logging the errors on backend will help to identify any future actions that we need to do in case of errors.
At the moment the websocket is a good add one - if the application works fine without having the socket connections then the 'overlay' is not required either at this point.

All of the points are valid though and we should reconsider as we add more things to the socket.

beniaminmunteanu:
I agree with you guys.

packages/wallet/frontend/src/pages/_app.tsx
```diff
@@ -13,6 +19,60 @@ const titilium = Titillium_Web({
...
+      withCredentials: true
+    })
```
beniaminmunteanu, line 29:
```suggestion
    socket = io(process.env.NEXT_PUBLIC_BACKEND_URL ?? '', {
      withCredentials: true,
      transports: ['websocket', 'polling'],
    })
```

packages/wallet/frontend/src/pages/_app.tsx
```diff
@@ -13,6 +19,60 @@ const titilium = Titillium_Web({
...
+      )
+    })
+
+    // log out the user when connection error occurs
+    socket.on('connect_error', async () => {
+      setTimeout(() => {
+        socket?.connect()
+      }, 1000)
+    })
+
```
beniaminmunteanu, line 69: This would be good to have on rafiki.money as well:

```suggestion
socket.on('reconnect_error', (error) => {
  console.error('Reconnection error:', error);
});
```

beniaminmunteanu (dismissed):

packages/boutique/frontend/src/components/ui/popover.tsx
raducristianpopa, line 1: Why was this file changed?

packages/boutique/frontend/src/components/ui/sheet.tsx
raducristianpopa, line 1: Ditto above.

packages/wallet/backend/src/rafiki/service.ts
```diff
@@ -233,6 +237,13 @@ export class RafikiService implements IRafikiService {
       { status: 'COMPLETED', value: amount.value }
     )
 
+    const user = await this.deps.userService.getByWalletId(receiverWalletId)
+    if (user)
+      await this.deps.socketService.emitMoneyReceivedByUserId(
+        user.id.toString(),
+        amount.assetCode
```
raducristianpopa, line 244: We should send the full amount object, so we can specify how much money the user received.

packages/wallet/frontend/src/lib/balance.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const balanceState = valtioPersist<AccountState>(state)
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function valtioPersist<TState extends Object>(
+  initialState?: TState
+): TState {
+  return proxy(initialState)
+}
```
raducristianpopa, line 24:
```suggestion
export const balanceState = proxy(state)
```

packages/wallet/frontend/src/pages/_app.tsx
```diff
@@ -13,6 +19,66 @@ const titilium = Titillium_Web({
...
+      transports: ['websocket', 'polling']
+    })
+
+    // Event listeners
+    socket?.on('MONEY_RECEIVED', (data) => {
+      toast({
+        description: (
+          <p>
+            <MoneyBird className="mr-2 inline-flex h-8 w-8 items-center justify-center" />
+            You received some {data.assetCode} into account {data.name}.
```
raducristianpopa, line 38: Let's specify how much money the user received.

packages/wallet/frontend/tailwind.config.js
```diff
@@ -66,5 +66,5 @@ module.exports = {
       }
     }
   },
-  plugins: [require('@tailwindcss/forms')]
+  plugins: [require('@tailwindcss/forms'), require('tailwindcss-animate')]
```
raducristianpopa, line 69: `tailwindcss-animate` needs to be installed as well.

```
["ERROR" - 08:21:34] Cannot find module 'tailwindcss-animate'
```

packages/wallet/frontend/src/pages/transfer/send.tsx
```diff
@@ -69,6 +69,7 @@ const SendPage: NextPageWithLayout<SendProps> = ({ accounts }) => {
     const selectedAccount = accounts.find(
       (account) => account.value === accountId
     )
+
     setBalance(
```
raducristianpopa, line 73: Should the real-time balance update also happen here?

packages/boutique/frontend/src/components/ui/popover.tsx
Tymmmy: Sorry, probably a bad merge from master.

beniaminmunteanu (approved):
backend-wise

raducristianpopa (approved):

PR #640, opened 2023-08-22
feat(wallet-frontend): Display converted amount with correct exchange rate in sending money

### LINKED ISSUE #536: Display the converted amount for cross-currency transactions

Tymmmy:
### Context

- fixes #536


raducristianpopa (commented):
Looks good, but there is a little bug I'm noticing. When pasting an incoming payment URL the `Amount` field is not populated. It should populate with the converted amount in a cross-currency situation or with the exact amount in a same currency situation.

raducristianpopa (approved):

PR #664, opened 2023-08-28
feat: implement exchange button feature

### LINKED ISSUE #442: Implement Exchange button feature
We have the exchange button in testnet, that should allow user to move/exchange money from one account to another, no matter the currency.

If the exchanged currency account does not exist, we need to create one.

Tymmmy:
### Context

- fixes #442


Tymmmy:
For now, we can't make the fees to be less for exchange then for SEND money, so the main difference will be that the Exchange will automatically create accound and PP if it doesn't exist. There will be a new story for removing the notification the user gets when exchanging money.

dragosp1011 (approved):

PR #668, opened 2023-08-29
feat: add Exchange backend endpoints

Andrei199991:
# NOTE 
For now, "exchange" will be an transfer between 1 account to 1 other account (**request** quote -> accept quote)

per Radu's brilliant suggestion, we will make a new PR where we use **transferLiquidity**, soon™.

#442

### DIFF
--- a/packages/wallet/backend/src/quote/service.ts
+++ b/packages/wallet/backend/src/quote/service.ts
@@ -5,11 +5,13 @@ import { PaymentPointer } from '@/paymentPointer/model'
 import { Asset, Quote } from '@/rafiki/backend/generated/graphql'
 import { RafikiClient } from '@/rafiki/rafiki-client'
 import { incomingPaymentRegexp, urlToPaymentPointer } from '@/utils/helpers'
-import { PaymentPointerService } from '../paymentPointer/service'
+import {
+  createPaymentPointerIfFalsy,
+  PaymentPointerService
+} from '../paymentPointer/service'
 import { QuoteWithFees } from './controller'
 import { RatesService } from '../rates/service'
 import { Account } from '@/account/model'
-import { getRandomValues } from 'crypto'
 
 type CreateExchangeQuote = {
   userId: string
@@ -46,31 +48,17 @@ type ConvertParams = {
   amount: bigint
 }
 
-const createPaymentPointerIfFalsy = async ({
-  paymentPointer,
-  userId,
-  accountId,
-  publicName,
-  paymentPointerService
-}: {
-  paymentPointer: PaymentPointer
-  userId: string
-  accountId: string
-  publicName: string
-  paymentPointerService: PaymentPointerService
-}): Promise<PaymentPointer> => {
-  if (paymentPointer) {
-    return paymentPointer
-  }
+const getAccountWithPaymentPointerBy = async (where: Partial<Account>) => {
+  const account = await Account.query()
+    .where(where)
+    .withGraphFetched({ paymentPointers: true })
+    .modifyGraph('paymentPointers', (builder) => {
+      builder.where({ active: true }).orderBy('createdAt', 'ASC').limit(1)
+    })
+    .limit(1)
+    .first()
 
-  const newPaymentPointer = await paymentPointerService.create(
-    userId,
-    accountId,
-    getRandomValues(new Uint32Array(1))[0].toString(16),
-    publicName
-  )
-
-  return newPaymentPointer
+  return account
 }
 
 export class QuoteService implements IQuoteService {
@@ -218,13 +206,10 @@ export class QuoteService implements IQuoteService {
     assetCode,
     amount
   }: CreateExchangeQuote): Promise<QuoteWithFees> {
-    const accountFrom = await Account.query()
-      .findById(accountId)
-      .where('userId', userId)
-      .withGraphFetched({ paymentPointers: true })
-      .modifyGraph('paymentPointers', (builder) => {
-        builder.where({ active: true }).orderBy('createdAt', 'ASC').limit(1)
-      })
+    const accountFrom = await getAccountWithPaymentPointerBy({
+      userId,
+      id: accountId
+    })
 
     if (!accountFrom) {
       throw new NotFound(`The source account does not exist for this user.`)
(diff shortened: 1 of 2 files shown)


packages/wallet/backend/src/account/service.ts
```diff
@@ -295,4 +308,84 @@ export class AccountService implements IAccountService {
...
+    const assetInRafiki = (
+      await this.deps.rafiki.listAssets({ first: 100 })
+    ).find((asset) => asset.code === assetCode)
+    if (!assetInRafiki) {
+      throw new NotFound(
+        `Asset Code "${assetCode}" does not exist in Rafiki; Payment Pointer could not be automatically created.`
+      )
+    }
+
+    let senderPpId = accountFrom?.paymentPointers?.[0]?.id
```
beniaminmunteanu, line 338: optionalChaining might  not be needed for accountFrom

packages/wallet/backend/src/account/service.ts
```diff
@@ -295,4 +308,84 @@ export class AccountService implements IAccountService {
...
+
+    let accountTo = await Account.query()
+      .where({ userId, assetCode })
+      .withGraphFetched({ paymentPointers: true })
+      .modifyGraph('paymentPointers', (builder) => {
+        builder.where({ active: true }).orderBy('createdAt', 'ASC').limit(1)
+      })
+      .limit(1)
+      .first()
+    let receiverPp = accountTo?.paymentPointers?.[0]?.url
```
beniaminmunteanu, line 359: move under the !accountTo check and remove optional chaining?

packages/wallet/backend/src/account/service.ts
```diff
@@ -31,10 +35,19 @@ interface IAccountService {
...
+  accountId: string
+  assetCode: string
+  amount: number
+}
+
 interface CountriesServiceDependencies {
   rapyd: RapydClient
   rafiki: RafikiClient
   logger: Logger
+  quoteServiceGetter: () => QuoteService
```
beniaminmunteanu, line 49: keep it just as quoteService, we don't use this naming for dependencies anywhere else

packages/wallet/backend/src/createContainer.ts
```diff
@@ -289,10 +294,9 @@ export const createContainer = (config: Env): Container<Bindings> => {
     return new RafikiController({ logger, rafikiService, ratesService })
   })
 
-  container.singleton(
-    'quoteService',
-    async () =>
-      new QuoteService({
+  container.singleton('quoteService', async () => {
+    if (!quoteServiceResolved) {
```
beniaminmunteanu, line 298: Imo I don't think this layer of caching services is needed. If you look inside the container class, everything is already kept inside a map, and fetching out of it is syncrhonous in itself. Just the container.resolve method itself is async, in case async dependencies are ever needed.

Same goes for above

packages/wallet/backend/src/account/controller.ts
```diff
@@ -131,4 +138,29 @@ export class AccountController implements IAccountController {
       next(e)
     }
   }
+
+  createExchangeQuote = async (
```
dragosp1011, line 142: define the method in interface

packages/wallet/backend/src/account/service.ts
```diff
@@ -295,4 +308,84 @@ export class AccountService implements IAccountService {
...
     return account
   }
+
+  public async createExchangeQuote({
+    userId,
+    accountId,
+    assetCode,
+    amount
+  }: CreateExchangeQuote): Promise<QuoteWithFees> {
+    const accountFrom = await Account.query()
```
dragosp1011, line 318: there should be a validation that the account belongs to current user

packages/wallet/backend/src/account/service.ts
```diff
@@ -295,4 +308,84 @@ export class AccountService implements IAccountService {
...
+      })
+
+    if (!accountFrom) {
+      throw new NotFound(`The source account does not exist.`)
+    }
+
+    const assetInRafiki = (
+      await this.deps.rafiki.listAssets({ first: 100 })
+    ).find((asset) => asset.code === assetCode)
+    if (!assetInRafiki) {
```
dragosp1011, line 332: shouldn't the same validation happen for receiver asset?

packages/wallet/backend/src/account/service.ts
```diff
@@ -295,4 +308,84 @@ export class AccountService implements IAccountService {
...
+      .withGraphFetched({ paymentPointers: true })
+      .modifyGraph('paymentPointers', (builder) => {
+        builder.where({ active: true }).orderBy('createdAt', 'ASC').limit(1)
+      })
+
+    if (!accountFrom) {
+      throw new NotFound(`The source account does not exist.`)
+    }
+
+    const assetInRafiki = (
```
dragosp1011, line 329: please extract the asset validation in a separate method

dragosp1011: as this may be needed in different services, probably the best place to be is in rafiki service

packages/wallet/backend/src/account/service.ts
```diff
@@ -295,4 +308,84 @@ export class AccountService implements IAccountService {
 
     return account
   }
+
+  public async createExchangeQuote({
```
dragosp1011, line 312: as a lot of the logic repeats for sender and receiver, I suggest to try to extract the logic that can be reused for both in multiple small methods

packages/wallet/backend/src/createContainer.ts
beniaminmunteanu: so this is a circular dependency case?

packages/wallet/backend/src/quote/service.ts
```diff
@@ -36,6 +46,33 @@ type ConvertParams = {
   amount: bigint
 }
 
+const createPpIfFalsy = async ({
```
dragosp1011, line 49: what is Pp? - better to use the full name IMO

packages/wallet/backend/src/quote/service.ts
```diff
@@ -174,4 +211,76 @@ export class QuoteService implements IQuoteService {
...
     )
   }
+
+  public async createExchangeQuote({
+    userId,
+    accountId,
+    assetCode,
+    amount
+  }: CreateExchangeQuote): Promise<QuoteWithFees> {
+    const accountFrom = await Account.query()
```
dragosp1011, line 221: how about moving this complex query into account service and make it more generic so in can be used in both cases
```
async getAccountWithPaymentPointerBy(where: Partial<Account>) {
    const account = await Account.query()
        .where(where)
        .withGraphFetched({ paymentPointers: true })
        .modifyGraph('paymentPointers', (builder) => {
          builder.where({ active: true }).orderBy('createdAt', 'ASC').limit(1)
        }).limit(1)
        .first()

    return account
  }
  ```
  
  and here
  ```
  const accountFrom = await this.deps.accountService.getAccountWithPaymentPointerBy({ id: accountId, userId })
  let accountTo = await this.deps.accountService.getAccountWithPaymentPointerBy({ userId, assetCode })
  ```

File: packages/wallet/backend/src/quote/service.ts [hunk 12]
```diff
@@ -36,6 +46,33 @@ type ConvertParams = {
   amount: bigint
 }
 
+const createPaymentPointerIfFalsy = async ({
```
dragosp1011, line 49: move to payment pointer service

dragosp1011 (approved):

PR #723, opened 2023-09-08
feat(commerce/backend): add pre-interaction checkout flow

### LINKED ISSUE #722: Add OP client

### LINKED ISSUE #726: Add first phase of the checkout flow (pre-interaction)

raducristianpopa:
### Context

- closes #722 
- closes #726


packages/commerce/backend/src/config/env.ts
```diff
@@ -3,9 +3,18 @@ import { z } from 'zod'
...
-    .default('postgres://postgres:password@localhost:5433/commerce_backend')
+    .default('postgres://postgres:password@localhost:5433/commerce_backend'),
+  AUTH_SERVER: z.string().url().default('http://rafiki-auth:3006'),
+  PAYMENT_POINTER: z.string().default('http://rafiki-backend/shop'),
+  KEY_ID: z.string().default('8e5e109e-9ea8-44a0-9490-1792bf1e07cb'),
+  PRIVATE_KEY: z
+    .string()
+    .default(
+      'LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1DNENBUUF3QlFZREsyVndCQ0lFSU9LemxweFR5c0xhbS94VWZ5R2t2T2dpSmcraTEvT25STURWMElqNi9vVnQKLS0tLS1FTkQgUFJJVkFURSBLRVktLS0tLQ=='
+    )
```
raducristianpopa, line 17: We will need to seed our wallet with a default account with two payment pointers. Will do this in a separate PR later on.

packages/commerce/backend/src/open-payments/service.ts
```diff
@@ -0,0 +1,202 @@
...
+                receiveAmount
+              }
+            }
+          ]
+        },
+        interact: {
+          start: ['redirect'],
+          finish: {
+            method: 'redirect',
+            uri: `${this.env.FRONTEND_URL}/placeholder?orderId=${orderId}`,
```
raducristianpopa, line 166: Placeholder URL for the moment.

.gitignore
```diff
@@ -133,3 +133,4 @@ dist
 # Others
 .DS_Store
 .idea
+tmp
```
beniaminmunteanu, line 136: what will be in the tmp directory?

packages/commerce/backend/src/cache/in-memory.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  async get(key: string): Promise<TData | null> {
+    const cached = this.cache.get(key)
+
+    if (!cached) {
+      return null
+    }
+
+    if (cached.expires !== null && cached.expires < Date.now()) {
```
beniaminmunteanu, line 27: Since this is a base cache class and might be used in other places as well directly as an abstraction, do you intend to also cleanup the cache on different instances except for when the specific item is retrieved and found as expired?

How about we also have a loop that periodically checks expired items? (Unless we want to have a notion of observability on cache misses which i don't think is the case)

What do you think about either a loop that checks expired items, or an implementation of an LRU cache?

raducristianpopa: I don't see how an LRU cache fits our case? Regarding the loop that periodically checks expired items, that's what I'm planning to do (we cache some order related fields so we won't need to save them in the database), but this is going to be in the second phase of the checkout flow (post-interaction).

.gitignore
raducristianpopa: Nothing at the moment. I stored some private keys while integrating the OP client and I've added it to `gitignore`.

packages/commerce/backend/src/open-payments/service.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    if (isPendingGrant(grant)) {
+      this.logger.error('Expected non-interactive quote grant.')
+      throw new InternalServerError()
+    }
+
+    return grant
+  }
+
+  private async createOutgointPaymentGrant(
```
dragosp1011, line 141: misspelling

packages/commerce/backend/src/config/env.ts
dragosp1011: why 2 payment pointers?

raducristianpopa: One that will be the shop's payment pointer and one that will act as the customer.

dragosp1011: I don't think the customer one is necessary as any payment pointer can be used for customer actions

packages/commerce/backend/src/cache/token.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  private cacheToken(token: AccessToken): void {
+    this.set(
+      'accessToken',
+      token.access_token.value,
+      token.access_token.expires_in
+    )
+
+    /**
+     * If we chage AUTH_SERVER_DOMAIN environment variable to
```
dragosp1011, line 71: typo on change

packages/commerce/backend/src/open-payments/service.ts
```diff
@@ -0,0 +1,202 @@
...
+  PendingGrant,
+  Quote,
+  isPendingGrant
+} from '@interledger/open-payments'
+import { randomUUID } from 'crypto'
+import { Logger } from 'winston'
+
+interface PreparePaymentParams {
+  order: Order
+  paymentPointer: string
```
dragosp1011, line 19: rename to `paymentPointerUrl` maybe, so it will be less confusing where is a mix of payment pointer objects and urls

dragosp1011 (dismissed):

dragosp1011 (approved):

PR #730, opened 2023-09-11
fix(rapyd): add additional beneficiary fields on withdraw money

### LINKED ISSUE #674: WIthdraw money works only for EUR and USD
Create an account of GBP or SGD, or any other account then EUR or USD.
WIthdraw funds gives error. Probably something to do with Raoyd side.
We should investigate

dragosp1011:
### Context

- fixes #674 

### Changes
- complete additional beneficiary and sender fields with fake data


Tymmmy:
Every currency has a different list of required beneficiary and sender fields. So adding some new fields for beneficiary will not cover all cases. We should get the list of required fields, and generate values with the given regexp in the code. Also looks like Rapyd SWIFT code has some issues that I opened a ticket with.

packages/wallet/backend/src/rapyd/rapyd-client.ts
```diff
@@ -179,19 +179,29 @@ export class RapydClient implements IRapydClient {
...
-      address: args.user.address ?? ''
+      first_name: args.user.firstName,
+      last_name: args.user.lastName,
+      address: args.user.address ?? '',
+      payment_type: 'priority', // https://docs.rapyd.net/en/payout-with-fx.html
+      city,
+      iban: 'DE75512108001245126199',
+      bic_swift: `BARC${payoutType.beneficiary_country.toUpperCase()}22`, // https://docs.rapyd.net/en/bic-swift-numbers-for-testing.html
+      date_of_birth: '22/02/1980'
     }
```
Tymmmy, line 203: Not enough, some currencies have specific required field

Tymmmy (approved):

Tymmmy (approved):

PR #758, opened 2023-09-19
feat(wallet-frontend): grant interaction fe improvement

### LINKED ISSUE #748: Grant interaction FE improvement
When you shop at rafiki.boutique and do a checkout, the Interaction page should be more user friendly. SO the user understands that the shop wants access to get the payment.

Tymmmy:
### Context

- fixes #748 

### Changes
User is prompted a simple question about the grant interaction (probably needs to be changed a bit, when the one click pay subscription is in place).
If the user is logged in then the interaction goes on without issues, if the user is not logged in, the user will be asked to log in, then back to interaction page.


packages/wallet/frontend/src/middleware.ts
```diff
@@ -52,9 +51,13 @@ export async function middleware(req: NextRequest) {
...
     // If the user is not logged in and tries to access a private resource,
-    // redirect to auth page.
+    // redirect to auth page or in the case of grant-interaction, back to the interaction page.
+    const callbackUrl =
+      req.url.indexOf('grant-interact') !== -1 ? `?callbackUrl=${req.url}` : ''
     if (!isPublic && !response.success) {
-      return NextResponse.redirect(new URL('/auth', req.url))
+      return NextResponse.redirect(
+        new URL(`/auth/login/${callbackUrl}`, req.url)
+      )
```
raducristianpopa, line 60: Instead of doing it only for the interaction page we should do it for every private routes.

raducristianpopa (approved):

PR #765, opened 2023-09-20
feat(payments): create external receiver

### LINKED ISSUE #764: Support external payments
Currently if there is a payment pointer - an incoming payment is created, if not - an error is returned.
Testnet should support external payments - so in case the payment pointer is not local, an external receiver should be created

dragosp1011:
### Context

- fixes #764

### DIFF
--- a/packages/wallet/backend/src/rafiki/backend/generated/graphql.ts
+++ b/packages/wallet/backend/src/rafiki/backend/generated/graphql.ts
@@ -96,6 +96,13 @@ export type AssetsConnection = {
   pageInfo: PageInfo;
 };
 
+export type BasePayment = {
+  createdAt: Scalars['String']['output'];
+  id: Scalars['ID']['output'];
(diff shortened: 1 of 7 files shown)


packages/wallet/backend/src/rafiki/backend/generated/graphql.ts
```diff
@@ -763,6 +803,11 @@ export type PaymentPointersConnection = {
   pageInfo: PageInfo;
 };
 
+export enum PaymentType {
+  Incoming = 'INCOMING',
```
rico191013, line 807: Well, I think UPPERCASE is better? like INCOMING = 'INCOMING'

dragosp1011: this file is generated from rafiki graphql schema

rico191013 (approved):

PR #818, opened 2023-09-28
fix(payment): continue payment if it is not internal

dragosp1011:
### Context

- fixes #

### DIFF
--- a/packages/wallet/backend/src/outgoingPayment/service.ts
+++ b/packages/wallet/backend/src/outgoingPayment/service.ts
@@ -17,11 +17,16 @@ export class OutgoingPaymentService implements IOutgoingPaymentService {
     const quote = await this.deps.rafikiClient.getQuote(quoteId)
     const paymentPointerId = quote.paymentPointerId
 
-    const incomingPayment =
-      await this.deps.incomingPaymentService.getPaymentDetailsByUrl(
-        quote.receiver
-      )
-    const { description } = incomingPayment
+    let description
+    try {
+      const incomingPayment =
+        await this.deps.incomingPaymentService.getPaymentDetailsByUrl(
+          quote.receiver
+        )
+      description = incomingPayment.description
+    } catch (_e) {
+      // @todo: find another way to get payment description
+    }
 
     await this.deps.rafikiClient.createOutgoingPayment({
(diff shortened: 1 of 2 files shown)


raducristianpopa:
Why is this change needed?

Tymmmy:
@raducristianpopa Transactions were not working on rafiki.money at all on friday and I think this is the fix....
@dragosp1011 correct me if I am wrong :D

Tymmmy (approved):

packages/wallet/backend/src/utils/helpers.ts
```diff
@@ -16,7 +16,7 @@ export const transformAmount = (
 }
 
 export const urlToPaymentPointer = (url: string) =>
-  url.split('/').slice(0, 4).join('/')
+  url.split('/incoming-payments/')[0]
```
dragosp1011, line 19: @raducristianpopa as in local playground in rafiki the payment pointers are `accounts/gfranklin`
when extracting the payment pointer from an incoming payment it was creating it as `accounts` only
this change should extract the full payment pointer

packages/wallet/backend/src/outgoingPayment/service.ts
```diff
@@ -17,11 +17,16 @@ export class OutgoingPaymentService implements IOutgoingPaymentService {
...
-    const { description } = incomingPayment
+    let description
+    try {
+      const incomingPayment =
+        await this.deps.incomingPaymentService.getPaymentDetailsByUrl(
+          quote.receiver
+        )
+      description = incomingPayment.description
+    } catch (_e) {
+      // @todo: find another way to get payment description
```
dragosp1011, line 28: @raducristianpopa `getPaymentDetailsByUrl` searches for a local incoming payment and if it does not exist it throws 404
in case of external payments there is no local incoming payment. The external outgoing payments won't have a description for now.

packages/wallet/backend/src/utils/helpers.ts
raducristianpopa: I think this is what we are looking for: 
- https://github.com/interledger/rafiki/issues/1005

raducristianpopa (approved):

PR #819, opened 2023-09-29
feat(wallet-frontend): Feedback changes for texts

### LINKED ISSUE #813: Onboarding flow and Grant access text improvements
Improve the text of the onboarding flow and the grant access infos.
They are a bit confusing for a first time user, so we need to make sure we are as clear as possible.

Tymmmy:
### Context

- fixes #813

### DIFF
--- a/packages/wallet/frontend/src/pages/grant-interactions/index.tsx
+++ b/packages/wallet/frontend/src/pages/grant-interactions/index.tsx
@@ -10,6 +10,7 @@ import Image from 'next/image'
 import { useDialog } from '@/lib/hooks/useDialog'
 import { ErrorDialog } from '@/components/dialogs/ErrorDialog'
 import { SuccessDialog } from '@/components/dialogs/SuccessDialog'
+import { useRouter } from 'next/router'
 
 type GrantInteractionPageProps = InferGetServerSidePropsType<
   typeof getServerSideProps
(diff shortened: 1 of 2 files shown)


packages/wallet/frontend/src/pages/grant-interactions/index.tsx
```diff
@@ -49,9 +51,14 @@ const GrantInteractionPage = ({
     openDialog(
       <SuccessDialog
         onClose={closeDialog}
+        onSuccess={() => {
+          router.push(
+            `${process.env.NEXT_PUBLIC_AUTH_HOST}/interact/${interactionId}/${nonce}/finish`
+          )
+          closeDialog()
+        }}
```
raducristianpopa, line 59: Why don't we automatically redirect the user after a few seconds instead of letting them click the `Finish` button?

Tymmmy: There is the finish api call from interaction with grants, we want to emphasise that.
It is definetly better to have this one button then also Close :D

raducristianpopa: I think the `finish` action is just an implementation detail. The user should *always* be redirected after interacting with a grant.

raducristianpopa (approved):

PR #839, opened 2023-10-05
Add back navigation and breadcrumb

### LINKED ISSUE #474: Back navigation
We should implement a back navigation for our pages.
This should be a small arrow representing "Go Back" <-

For example we have the pages Accounts->Payment Pointers->Transactions, and maybe a breadcrumb navigation would help here the users. We can also think on having them on all pages and the back button bringing us back to home page. This would be especially helpful for mobile users, where the main menu is not visible all the time.

The back <= should appear on the very top left side.

umeshSinghVerma:
## Issue
fixes #474 

## Description

In this pull request, I've implemented two key features to enhance the user experience:

1. **Back Navigation Button**: Added a back navigation button to the user interface. This button allows users to easily return to the home page from any point in the application, improving overall navigation.

2. **Breadcrumbs**: Implemented breadcrumbs at the top of the page. Breadcrumbs provide users with a clear and intuitive trail of their navigation history. Users can now see their path and easily backtrack to previous pages, which is particularly useful for deep navigation within the application.

## Changes Made

- Created a back navigation button component with back arrow icon, making it highly visible and accessible to users.
- Integrated the back navigation functionality, ensuring it redirects users to the home page.
- Introduced breadcrumb elements at the top of each page, dynamically generated based on the user's navigation path.
- Styled the breadcrumbs for consistency with the application's design.

## Screenshots
Desktop view:
image

Mobile View
image

### DIFF
--- a/packages/wallet/frontend/src/components/Menu.tsx
+++ b/packages/wallet/frontend/src/components/Menu.tsx
@@ -196,10 +196,31 @@ export const Menu = () => {
       <nav className="fixed inset-x-0 z-10 flex h-20 flex-col bg-white shadow-md md:inset-y-0 md:h-auto md:w-60 md:shadow-none">
         <div className="flex min-h-0 flex-1 items-center px-6 py-10 md:flex-col md:items-start md:overflow-y-auto md:bg-gradient-primary">
           <div className="flex items-center font-semibold text-green">
-            <button className='w-4 h-4 scale-125 md:fixed md:top-3 md:left-2 flex items-center mr-5 mt-[5px] active:scale-100' onClick={() => {
-              router.push('/');
-            }}>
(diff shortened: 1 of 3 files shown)


Tymmmy:
@umeshSinghVerma thank you for the PR, however I have a few things I would like you to change.
Let's think about user experience every time, and that means we should'n do things that are redundant, or duplicated, also we shouldn't confuse users more then necesary :) .
What do I mean by this?

1. "Back Navigation Button: Added a back navigation button to the user interface. This button allows users to easily return to the home page from any point in the application, improving overall navigation." You implemented the back button with the icon back, that brings you to HOME page in the area of the main menu. For this we already have the HOME link from the main menu. We don't need another one in this area, because we already have one. I would suggest instead to make the Interledger Testnet and Logo Duo a link to home page as well, that way mobile will also have a direct way to go HOME from here as well. 
       - So please remove the back button from the main menu area, and make the Logo and text a link to take you to HOME:
image

2. The breadcrumbs.... That is a good idea, BUT the user would get confused when they see this:
image
They can ask what is that long list of numbers and characters? 
And for example on Developer Keys we have 
image
If you click on Settings from the breadcrumb it brings you to the personal settings page, and if the user went to Developer settings directly from main menu, then this is a bug.
Then go to Send page or Request page from Transfers:
image
Click on Transfer breadcrumb link -> Error 404 page so another BUG
And when you are on Home page already we have a breadcrumb with Home again...duplicate and unnecesarry link to Home page, when you are already on Home page.
   - So my suggestion here is to rethink the breadcrumb....don't display everything you see in the url. Display here the arrow button (add a separate file for the svg and just add for example <BackButton />), make sure the pages that are one level bellow Home page takes you back to Home page with clicking on this button. Then for the navigation of Home page - > select an Account -> and click on View from a payment Pointer we can have a longer breadcrumb of Home->Account. And when you click on Account it takes you back to the page of the previously selected Account with the list of payment pointers. (use the same text color as we have the Menu links, the dark green) 


Hope I wrote everything clear enough.

Tymmmy (changes requested):
Remove icon back from menu area, rething breadcrumbs (see detailed explanation in comment)

umeshSinghVerma:
Hello @Tymmmy,

Thank you for your valuable guidance and insights. I have comprehended all the aspects discussed, and I am now prepared to conclude this Pull Request. I will proceed to generate a new Pull Request that incorporates all the modifications and improvements you have recommended.

PR #862, opened 2023-10-10
fix: set the value from incoming payment in quote

### LINKED ISSUE #861: Request payment error
Create an account with RON currency, and one with EUR currency.
Request 100 RON from the account, and try sending/paying it with the EUR account.
Even if you have enough money, and everything should be ok we have an error.
image

Same happens for HUF or DKK.

EUR USD GBP work.

dragosp1011:
### Context

- fixes #861

### Changes
- uses the amount from incoming payment instead of converting it
this PR will solve issues with internal cross-currencies payments.
External payments may still encounter this issue and it will be solved when https://github.com/interledger/rafiki/issues/1005 will be implemented.

### DIFF
--- a/packages/wallet/backend/src/quote/service.ts
+++ b/packages/wallet/backend/src/quote/service.ts
@@ -4,7 +4,11 @@ import { IncomingPaymentService } from '@/incomingPayment/service'
 import { PaymentPointer } from '@/paymentPointer/model'
 import { Asset, Quote } from '@/rafiki/backend/generated/graphql'
 import { RafikiClient } from '@/rafiki/rafiki-client'
-import { incomingPaymentRegexp, urlToPaymentPointer } from '@/utils/helpers'
+import {
+  incomingPaymentRegexp,
+  transformBalance,
+  urlToPaymentPointer
+} from '@/utils/helpers'
 import {
   createPaymentPointerIfFalsy,
   PaymentPointerService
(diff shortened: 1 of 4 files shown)


packages/wallet/backend/src/quote/service.ts
```diff
@@ -109,11 +113,27 @@ export class QuoteService implements IQuoteService {
       params.isReceive &&
       destinationPaymentPointer.assetCode !== asset.code
     ) {
-      const convertedValue = await this.convert({
-        from: assetCode,
-        to: destinationPaymentPointer.assetCode,
-        amount: value
-      })
+      let convertedValue
+      if (isIncomingPayment) {
```
raducristianpopa, line 117: I'm not sure about this approach. The users should be able send their desired amounts. If the incoming payment has a receive amount of 100EUR, we should not enforce to send 100EUR.

Tymmmy: Yep, I agree, the request might be 100EUR but the user ca decide to pay only 50EUR

dragosp1011: the issue is that converting the value is not exact, so 20.15 euro will result in 100.02 roni (instead of expected 100)
I added a check that if the amount is off by less then 1% to set it to the amount in incoming payment
open to other suggestions if this does not make sense

beniaminmunteanu (dismissed):

beniaminmunteanu (approved):

PR #866, opened 2023-10-11
feat(wallet-frontend): Add copy button for PP interledger#814

### LINKED ISSUE #814: Add a copy button next to the payment pointers in the list
This is for lazy people :D 
We could add a copy button next to the payment pointers name so people don't have to select and copy :D 
We already use the copy mechanism for Dev Keys in settings, we could add the same button next to the PP name inside ACcount details.

image

daniloff200:
If the PR is related to an open issue(s) please provide a list of them.
Closes #814

### Context
Added copy button for PP

### DIFF
--- a/packages/wallet/frontend/src/components/cards/PaymentPointerCard.tsx
+++ b/packages/wallet/frontend/src/components/cards/PaymentPointerCard.tsx
@@ -13,6 +13,7 @@ import { EllipsisV } from '../icons/Ellipsis'
 import { PencilSquare } from '../icons/Pencil'
 import { Trash } from '../icons/Trash'
 import { EditPaymentPointerDialog } from '../dialogs/EditPaymentPointerDialog'
+import { CopyButton } from '@/ui/CopyButton'
 
 type PaymentPointerCardProps = {
   paymentPointer: PaymentPointer
@@ -33,7 +34,7 @@ const PaymentPointerCardButton = forwardRef<
       ref={ref}
       className={cx(
         className,
-        'inline-flex items-center justify-center rounded-md px-2 font-medium text-white hover:shadow-md'
+        'mr-4 inline-flex items-center justify-center rounded-md px-2 font-medium text-white hover:shadow-md'
       )}
       onClick={() => {
         if (isUserFirstTime) {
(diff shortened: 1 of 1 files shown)


Tymmmy:
@daniloff200 a few comments here:
 - let's make the copy button a bit smaller, so the size of the row stays the same as it was before
 - add more then one PP to the list, you will se that the View button will not be aligned anymore, but all over the place, depending on the PP name size, so aligh to the right next to the copy button.
 
image

packages/wallet/frontend/src/components/cards/PaymentPointerCard.tsx
```diff
@@ -63,7 +64,13 @@ export const PaymentPointerCard = ({
         >
           View
         </PaymentPointerCardButton>
+        <CopyButton
+            aria-label="copy key id"
```
Tymmmy, line 68:
```suggestion
            aria-label="copy pp"
```

daniloff200:
hi, @Tymmmy 
I've made the button smaller, to make it fit to the row.
Does it looks good in this case?
Selection_001

Tymmmy:
Well... more like the inside as well to be sized down, and be more like a square, and ofcourse in on line (the current branch gives me the result in two rows)

image

Tymmmy:
Comparing previous version, and  new version you can see that even the View button got bigger with a few pixels. We don't want that button to be bigger. Also please see CopyButton.tsx, inside the ClipboardCheck and Clipboard tag is resized when screen is small. Here is the logic where you need to add something in order to make the inside of the copy button smaller.

image

Tymmmy (approved):

Tymmmy:
Great job and thank you @daniloff200,  approved and squashed-merged into master.
Please follow announcements for the swag details here: https://community.interledger.org/devcer/hacktoberfest-with-interledger-join-the-open-payments-movement-3bkd

devcer:
Thank you for Contributing to Interledger during Hacktoberfest. If you have any accepted PRs, You can claim your Swag [here](https://community.interledger.org/devcer/hacktoberfest-swag-redemption-claim-your-cool-gear-2152). Last date to fill up the form is December 10th, 2023.  @daniloff200

daniloff200:
Hi, @devcer !
Thanks for the update!
I have a question about the shipping - I'm currently located in Ukraine, and you most likely now about the military actions there, so, what I want to ask  - is the shipping in Ukraine will be possible? Previously there were some issues with that, and in this case I've asked to send the swag to my friends in other countries, and then they've send that to me

devcer:
Hi Dmitriy, We will be using https://www.bonfire.com/ for the swag
shipping. You can check with them if they are shipping it to Ukraine. If
they do then you can use the Ukraine address or else your friends. Reach
out to me on Interledger slack to talk further.

PR #889, opened 2023-10-16
feat(wallet-frontend): private key copy as base 64

### LINKED ISSUE #867: Base64 encoded private key
When I generate a private key on Rafiki.money, it displays it (and also downloads the key file).
image
Can we have a second option next to "copy private key" or even replace "copy private key" with "copy base64 encoded private key"? 
Context: the AWS lambda to generate signatures expects the private key as a base64 encoded string.

Tymmmy:
### Context

- fixes #867 

### Changes
Change Copy Private button, to copy private button as base 64 encoded string.

### DIFF
--- a/packages/wallet/frontend/src/components/settings/DeveloperKeys.tsx
+++ b/packages/wallet/frontend/src/components/settings/DeveloperKeys.tsx
@@ -285,10 +285,14 @@ const PaymentPointerCTA = () => {
                   value={privateKey}
                 />
                 <CopyButton
-                  ctaText="COPY BASE64 PRIVATE KEY"
-                  copyType="base64"
-                  aria-label="copy base64 private key"
+                  ctaText="COPY PRIVATE KEY"
+                  aria-label="copy private key"
                   value={privateKey}
+                />
+                <CopyButton
+                  ctaText="COPY BASE64 ENCODED PRIVATE KEY"
+                  aria-label="copy base64 encoded private key"
+                  value={btoa(privateKey.trim())}
                 />
               </div>
             </div>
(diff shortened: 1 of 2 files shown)


packages/wallet/frontend/src/components/settings/DeveloperKeys.tsx
```diff
@@ -285,8 +285,9 @@ const PaymentPointerCTA = () => {
                   value={privateKey}
                 />
                 <CopyButton
-                  ctaText="COPY PRIVATE KEY"
-                  aria-label="copy private key"
+                  ctaText="COPY BASE64 PRIVATE KEY"
+                  copyType="base64"
+                  aria-label="copy base64 private key"
```
raducristianpopa, line 290: Let's keep both buttons:
- `Copy private key`
- `Copy Base64 encoded private key` or something similar

packages/wallet/frontend/src/ui/CopyButton.tsx
```diff
@@ -3,20 +3,22 @@ import { useEffect, useState } from 'react'
...
-  navigator.clipboard.writeText(value)
+function copyToClipboard(value: string, copyType?: string) {
+  navigator.clipboard.writeText(copyType === 'base64' ? btoa(value.trim()) : value)
 }
 
 type CopyButtonProps = Omit<ButtonProps, 'intent'> & {
   value: string
   ctaText?: string
   afterCtaText?: string
+  copyType?: string
```
raducristianpopa: We can revert these changes and instead of calling `btoa` in the `copyToClipboard` function, we should the specify the base64 encoded key directly:
```jsx
<CopyButton
  // ...rest
  value={btoa(privateKey)}
/>
```

packages/wallet/frontend/src/ui/CopyButton.tsx
```diff
@@ -3,20 +3,22 @@ import { useEffect, useState } from 'react'
...
 }
 
 export const CopyButton = ({
   value,
   afterCtaText,
   ctaText,
+  copyType,
```
raducristianpopa, line 23: With the above changes, `copyType` is not needed anymore.
```suggestion
```

raducristianpopa (approved):

PR #890, opened 2023-10-16
feat: remove websocket notification for Exchange

### LINKED ISSUE #860: Remove websocket notification for Exchange currency
The user is able to exchange currencies between their own accounts with the Exchnage button.
image

In the background the same outgoing-payments and incomin-paymnet is created as for Sending money.
While we can't influence for now the fees for this, we would like the user to NOT receive notification that they received money.
This means we need to add to metadata a flag maybe, that the outcoming and incoming payments are from exchange button, and then when we fire the MONEY_RECEIVED for incoming payments, in this case we DON'T fire the event.

rico191013:
### Context

- fixes #860

### DIFF
--- a/packages/wallet/backend/src/quote/service.ts
+++ b/packages/wallet/backend/src/quote/service.ts
@@ -17,7 +17,7 @@ import {
 import { QuoteWithFees } from './controller'
 import { RatesService } from '../rates/service'
 import { Account } from '@/account/model'
-import { Node_Cache } from '@/utils/helpers'
+import { NodeCacheInstance } from '@/utils/helpers'
 type CreateExchangeQuote = {
   userId: string
   accountId: string
@@ -283,9 +283,9 @@ export class QuoteService implements IQuoteService {
     })
 
     //OUTCOME
-    Node_Cache.set(quote.id, quote)
+    NodeCacheInstance.set(quote.id, true)
     //INCOME
-    Node_Cache.set(urlToPaymentId(quote.receiver), quote)
+    NodeCacheInstance.set(urlToPaymentId(quote.receiver), true)
     return quote
   }
 }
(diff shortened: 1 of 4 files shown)


beniaminmunteanu (approved):

packages/wallet/backend/src/utils/helpers.ts
```diff
@@ -79,3 +82,7 @@ export const prefixSomeObjectKeys = (
     }),
     {}
   )
+
+import NodeCache from 'node-cache'
+
+export const Node_Cache = new NodeCache({ stdTTL: 60 * 60 })
```
dragosp1011, line 88: maybe rename to NodeCacheInstance

packages/wallet/backend/src/rafiki/service.ts
```diff
@@ -315,7 +317,8 @@ export class RafikiService implements IRafikiService {
     )
 
     const user = await this.deps.userService.getByWalletId(source_ewallet)
-    if (user)
+    const isExchange = Node_Cache.get(wh.data.payment.id)
```
dragosp1011, line 320: is the payment id equal with quote id?

packages/wallet/backend/src/quote/service.ts
```diff
@@ -255,10 +259,14 @@ export class QuoteService implements IQuoteService {
       paymentPointerId: senderPpId,
       amount,
       description: 'Currency exchange.',
-      isReceive: false,
+      isReceive: true,
       receiver: receiverPpUrl
     })
 
+    //OUTCOME
+    Node_Cache.set(quote.id, quote)
```
dragosp1011, line 267: as the quote is not used and only helps to check if it is an exchange payment, cache it with a `true` value (or empty object if it requires the value to be an object)

packages/wallet/backend/src/rafiki/service.ts
rico191013: It's a little weird, but according to logs and database records, the IDs of outgoing payments and quotes are the same.

dragosp1011 (approved):

Issue #880, opened 2023-10-13
[WM] WM Payment pointer module


raducristianpopa:
- ~~Redis cache for fast retrieval of WM payment pointers~~
- ~~Methods:~~
  - ~~get (redis + database)~~
  - ~~create~~
  - ~~update balance~~
- ~~In the main Payment Pointer service make use of the WMPP get method which should return something similar to:~~
WM Payment pointers are now a part of the main PP module.

PR #896, opened 2023-10-16
feat(wallet-backend): introduce web-monetization payment pointers

beniaminmunteanu:
implements: #880 

This PR introduces web monetization payment pointers. It also adds Caching over Redis.


packages/wallet/backend/src/paymentPointer/webMonetization/model.ts
raducristianpopa, line 1: Why do we need this new model? We already have the models:

https://github.com/interledger/testnet/blob/hackathon/packages/wallet/backend/src/webMonetization/paymentPointer/model.ts

I forgot to add `assetCode` and `assetScale` to the initial `WMPaymentPointerModel`.

packages/wallet/backend/src/createContainer.ts
```diff
@@ -198,7 +221,8 @@ export const createContainer = (config: Env): Container<Bindings> => {
       new PaymentPointerService({
         env: await container.resolve('env'),
         rafikiClient: await container.resolve('rafikiClient'),
-        accountService: await container.resolve('accountService')
+        accountService: await container.resolve('accountService'),
+        wmPaymentPointerService: await container.resolve('wmPaymentPointerService')
```
raducristianpopa, line 225: Why does Rapyd need `wmPaymentPointerService`?

packages/wallet/backend/src/paymentPointer/webMonetization/service.ts
```diff
@@ -0,0 +1,148 @@
...
+        this.deps.rafikiClient.updatePaymentPointer({
+          id: paymentPointerId,
+          publicName
+        })
+      ])
+      await trx.commit()
+    } catch (e) {
+      await trx.rollback()
+    }
+  }
```
raducristianpopa, line 142: I would say to only have the `updateBalance` method for now Let's not worry about updating the public name or other related fields.

packages/wallet/backend/src/paymentPointer/webMonetization/service.ts
raducristianpopa, line 1: We need the generate the keys as well. Maybe we can reuse the ones from the main payment pointer service?

docker/dev/docker-compose.yml
```diff
@@ -45,10 +45,11 @@ services:
       FROM_EMAIL: ${FROM_EMAIL}
       SEND_EMAIL: ${SEND_EMAIL}
       RATE_API_KEY: ${RATE_API_KEY}
-      BASE_ASSET_SCALE: 2,
-      MAX_ASSET_SCALE: 9,
+      BASE_ASSET_SCALE: 2
+      MAX_ASSET_SCALE: 9
```
raducristianpopa, line 49: Ops. Missed that. Sorry!

packages/wallet/backend/src/createContainer.ts
beniaminmunteanu: it doesn't, it's a leftover. Just used it to test redis interaction

packages/wallet/backend/src/paymentPointer/webMonetization/service.ts
```diff
@@ -0,0 +1,148 @@
...
+    
+    return wmPaymentPointer
+  }
+
+  async getById(
+    userId: string,
+    accountId: string,
+    id: string
+  ): Promise<WMPaymentPointer> {
+    const cacheHit = await this.deps.cache.get<WMPaymentPointer>(id);
```
dragosp1011, line 94: the method that will be used in webhooks should check the cache and request the db based on id only (with active: true check) - the check for account and user is not needed for this case

beniaminmunteanu: Thanks. I had already moved it out, have another look 🙏

packages/wallet/backend/src/cache/redis-client.ts
```diff
@@ -0,0 +1,48 @@
...
+  set<T>(
+    key: string,
+    value: T | string,
+    options?: EntryOptions
+  ): Promise<string>
+  get<T>(key: string): Promise<T | null>
+  delete(key: string): Promise<number>
+}
+
+export class RedisClient implements IRedisClient {
```
raducristianpopa, line 17: Add a generic so we won't need to specify the type every time.

packages/wallet/backend/src/cache/service.ts
```diff
@@ -0,0 +1,37 @@
+import { EntryOptions, IRedisClient } from './redis-client'
+
+export interface ICacheService {
+  set<T>(key: string, value: T | string): Promise<string>
+  get<T>(key: string): Promise<T | null>
+  delete(key: string): Promise<number>
+}
+
+export class CacheService implements ICacheService {
```
raducristianpopa, line 9: Ditto above.

packages/wallet/backend/src/createContainer.ts
```diff
@@ -179,6 +183,39 @@ export const createContainer = (config: Env): Container<Bindings> => {
     async () => new RatesService({ env: await container.resolve('env') })
   )
 
+  container.singleton('redisClient', async () => {
+    //config for redis connection is in the Redis constructor. Default to
```
raducristianpopa, line 187:
```suggestion
```

docker/dev/docker-compose.yml
```diff
@@ -45,10 +45,11 @@ services:
...
       WM_THRESHOLD: 100000000
       DEBT_THRESHOLD: 5
+      REDIS_URL: ${REDIS_URL}
```
raducristianpopa, line 52: Add this to the prod compose file and `.env` example file.

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -74,6 +78,8 @@ export const createPaymentPointerIfFalsy = async ({
   return newPaymentPointer
 }
 
+// export type PaymentPointerResponse = (WMPaymentPointer | PaymentPointer) & {isWM:boolean}
+// type PaymentPointerResponse = (WMPaymentPointer & {isWM:true}) | (PaymentPointer & {isWM: false})
```
raducristianpopa, line 82:
```suggestion
```

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -149,10 +155,19 @@ export class PaymentPointerService implements IPaymentPointerService {
     userId: string,
     accountId: string,
     id: string
-  ): Promise<PaymentPointer> {
+  ): Promise<PaymentPointer | WMPaymentPointer> {
     // Validate that account id belongs to current user
```
raducristianpopa, line 159:
```suggestion
```

packages/wallet/backend/src/webMonetization/paymentPointer/model.ts
```diff
@@ -5,8 +5,11 @@ import { WMTransaction } from '../transaction/model'
 
 export class WMPaymentPointer extends PaymentPointerBaseModel {
   static tableName = 'wmPaymentPointers'
+  public isWM = true
 
   public balance!: bigint
+  public assetCode!: string
+  public assetScale!: number
```
raducristianpopa, line 12:
```suggestion
  static tableName = 'wmPaymentPointers'
  
  public isWM = true
  public balance!: bigint
  public assetCode!: string
  public assetScale!: number
```

packages/wallet/backend/src/webMonetization/paymentPointer/service.ts
```diff
@@ -0,0 +1,127 @@
...
+  async create(
+    params: CreateWMPaymentPointerParams
+  ): Promise<WMPaymentPointer> {
+    const account = await this.deps.accountService.findAccountById(
+      params.accountId,
+      params.userId
+    )
+
+    const url = `${this.deps.env.OPEN_PAYMENTS_HOST}/${params.paymentPointerName}`
+    let wmPaymentPointer = await WMPaymentPointer.query().findOne({ url })
```
raducristianpopa, line 51: Before creating a payment pointer we need to check in both tables.

packages/wallet/backend/src/webMonetization/paymentPointer/model.ts
beniaminmunteanu: no longer needed, wmPaymentPointer model and table no longer exist

packages/wallet/backend/src/webMonetization/paymentPointer/service.ts
beniaminmunteanu: no longer works this way.

packages/wallet/backend/src/account/service.ts
```diff
@@ -131,10 +131,13 @@ export class AccountService implements IAccountService {
     let query = Account.query().where('userId', userId)
     if (hasPaymentPointer)
       query = query
-        .withGraphFetched({ paymentPointers: true })
+        .withGraphFetched({ paymentPointers: true, wmPaymentPointers: true })
```
dragosp1011, line 134: `wmPaymentPointers` should be removed

packages/wallet/backend/src/cache/service.ts
beniaminmunteanu: done

beniaminmunteanu:
Frontend part just concats wmPaymentPointers with normal paymentPointers now, so the code doesn't break. 
@raducristianpopa  @Tymmmy  maybe you decide further on how you want to aggregate/segregate the data.
But the list payment pointers request now returns 
`{
   wmPaymentPointers:[],
   paymentPointers:[]
}`
As per @raducristianpopa 's request

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -64,107 +81,175 @@ export const createPaymentPointerIfFalsy = async ({
...
-    return PaymentPointer.query().joinRelated('account').where({
-      'account.userId': userId,
-      'paymentPointers.active': true
-    })
+    return PaymentPointer.query()
+      .where({ isWM: false })
+      .joinRelated('account')
+      .where({
+        'account.userId': userId,
+        'paymentPointers.active': true
```
dragosp1011, line 206: should this check be moved above as well?

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -64,107 +81,175 @@ export const createPaymentPointerIfFalsy = async ({
...
+      if (args.isWM) {
+        //* Web monetization feature requires an asset with scale MAX_ASSET_SCALE to exist. It's default assetCode is USD for now
+        const webMonetizationAsset =
+          await this.deps.rafikiClient.getRafikiAsset(
+            'USD',
+            this.deps.env.MAX_ASSET_SCALE
+          )
+
+        if (!webMonetizationAsset) {
+          throw new NotFound('Web monetization asset not found')
```
Tymmmy, line 145:
```suggestion
          throw new NotFound('Web monetization asset not found.')
```

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -64,107 +81,175 @@ export const createPaymentPointerIfFalsy = async ({
...
+
     return paymentPointer
   }
 
+  async updateBalance(args: UpdatePaymentPointerBalanceArgs): Promise<void> {
+    const { accountId, paymentPointerId, balance } = args
+
+    const paymentPointer = await this.getById({ accountId, paymentPointerId })
+    if (!paymentPointer) {
+      throw new NotFound(`WM Payment pointer does not exist`)
```
Tymmmy, line 248:
```suggestion
      throw new NotFound(`Web monetization payment pointer does not exist.`)
```

packages/wallet/frontend/src/pages/account/[accountId].tsx
```diff
@@ -202,10 +202,12 @@ export const getServerSideProps: GetServerSideProps<{
     }
   }
 
-  const paymentPointers = paymentPointersResponse.data.map((pp) => ({
-    ...pp,
-    url: pp.url.replace('https://', '$')
-  }))
+  const paymentPointers = paymentPointersResponse.data.paymentPointers
+    .concat(paymentPointersResponse.data.wmPaymentPointers)
```
Comment (Tymmmy) [inline packages/wallet/frontend/src/pages/account/[accountId].tsx:206]: In the accounts page, why do we also show in the list of PPs the WM PPs? 
I am asking because there will be a separate list for WM PPs. So why do we show them in the general PP list as well?

Comment (beniaminmunteanu) [reply packages/wallet/frontend/src/pages/account/[accountId].tsx:206]: Read my comment above:
https://github.com/interledger/testnet/pull/896#issuecomment-1768372919

packages/wallet/backend/src/paymentPointer/service.ts
beniaminmunteanu: like this?

`return PaymentPointer.query()
      .where({ isWM: false, active: true })
      .joinRelated('account')
      .where({
        'account.userId': userId,
      })`

packages/wallet/frontend/src/pages/account/[accountId].tsx
Comment (beniaminmunteanu) [reply packages/wallet/frontend/src/pages/account/[accountId].tsx:206]: this is just so the frontend does not crash on the response, I am expecting the client code on what to do with the two lists to be in a separate PR

packages/wallet/backend/src/paymentPointer/service.ts
dragosp1011: yes

packages/wallet/frontend/src/pages/transfer/send.tsx
raducristianpopa, line 1: In both Send and Request pages, I think we should exclude the WM PP, therefore the `concat` is not needed. For actual transactions made through the UI, we should only allow the normal payment pointers for now.

packages/wallet/frontend/src/lib/api/paymentPointer.ts
```diff
@@ -40,6 +40,11 @@ export type PaymentPointer = {
   keyIds: PaymentPointerKey | null
 }
 
+export type ListPaymentPointersResult = {
+  wmPaymentPointers: Array<PaymentPointer>
+  paymentPointers: Array<PaymentPointer>
+}
```
raducristianpopa, line 46: `listAll` function needs a separate type for the response, since it's not returning `{ paymentPointers: [], wmPaymentPointers: [] }`.

packages/wallet/frontend/src/pages/account/[accountId].tsx
Comment (beniaminmunteanu) [reply packages/wallet/frontend/src/pages/account/[accountId].tsx:206]: @Tymmmy  do you want me to also remove the wmPaymentPointers here?

Tymmmy (approved):

raducristianpopa (approved):

dragosp1011 (approved):

PR #902, opened 2023-10-18
feat(wallet-frontend): Display wm payment pinters in separate list

### LINKED ISSUE #899: [WM] Display WM Payment Pointers in separate list
- add a new menu item - Web Monetization
- display in the new page a list of payment pointers used for WM
- the list will be similar to the Accounts 'normal' PPs for now

Tymmmy:
### Context

- fixes #899 

### Changes
 - added tabs to Account view - Account & Web Monetization
 - Account Tab will be the same as before
 - Web Monetization will display list of web monetization payment pointers


packages/wallet/frontend/src/components/cards/PaymentPointerCard.tsx
```diff
@@ -51,20 +60,28 @@ PaymentPointerCardButton.displayName = 'PaymentPointerCardButton'
 
 export const PaymentPointerCard = ({
   paymentPointer,
+  isWM,
```
raducristianpopa, line 63:
```suggestion
  isWM = false,
```

packages/wallet/frontend/src/components/cards/PaymentPointerCard.tsx
```diff
@@ -51,20 +60,28 @@ PaymentPointerCardButton.displayName = 'PaymentPointerCardButton'
...
+          ) : null}
+          {!isWM ? (
+            <PaymentPointerCardButton
+              href={`/transactions?paymentPointerId=${paymentPointer.id}`}
+              aria-label="view payment pointer"
+              id={idOnboarding}
+            >
+              View
+            </PaymentPointerCardButton>
+          ) : null}
```
raducristianpopa, line 84:
```suggestion
          {isWM ? (
            <span className="flex items-center justify-center px-3">
              {formattedAmount(paymentPointer).amount}
            </span>
          ) : (
            <PaymentPointerCardButton
              href={`/transactions?paymentPointerId=${paymentPointer.id}`}
              aria-label="view payment pointer"
              id={idOnboarding}
            >
              View
            </PaymentPointerCardButton>}
          }
```

packages/wallet/frontend/src/components/tabs/AccountTabs.tsx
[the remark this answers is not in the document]
Tymmmy: It was needed when I did it with 2 pages, now it would not be needed here exactly, but it still would be needed for Settings TAbs, in order to not repeat the code in the 2 pages (index & developer-keys) it is being used. So even if I remove from AccountTabs the tab options and add them to [accountID] page, the Tab and SettingsTab are still needed. So I would suggest in order to keep it neat and tidy to still have this next to the other Tabs.

packages/wallet/frontend/src/components/cards/PaymentPointerCard.tsx
Tymmmy: Not needed anymore, if I am always using same [accountId] page to come to the PPCards component, isWM will always have value

raducristianpopa (approved):
Fixing existing errors in #905.

PR #905, opened 2023-10-19
feat(wallet/backend): add wm payment pointer transactions worker

### LINKED ISSUE #904: [WM] Payment pointer transactions worker

raducristianpopa:
### Context

- fixes #904


packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -413,6 +426,136 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+    if (paymentPointer.isWM) {
+      await this.deps.cache.set(paymentPointer.id, paymentPointer, {
+        expiry: 60
+      })
+    }
+
     return paymentPointer
   }
+
+  async processWMPaymentPointers(): Promise<void> {
```
raducristianpopa, line 438: This is the only method that needs to be reviewed. The other changes are from Dragos' PR.

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -413,6 +426,136 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+              status: 'COMPLETED',
+              description: 'Web Monetization'
+            }),
+            paymentPointer.$query(trx).update({
+              incomingBalance: raw('?? - ?', [
+                'incomingBalance',
+                this.deps.env.WM_THRESHOLD * incomingBalance
+              ])
+            })
+          ])
```
raducristianpopa, line 516: Will be moved in a separate method.

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -144,20 +151,38 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+          )
+
+        paymentPointer = await PaymentPointer.query().insert({
+          url: rafikiPaymentPointer.url,
+          publicName: args.publicName,
+          accountId: args.accountId,
+          id: rafikiPaymentPointer.id,
+          isWM: args.isWM
+        })
+      }
```
raducristianpopa, line 185: Any suggestions on how we can make this more readable? I had to split it like this because the WM payment pointer was getting created with the original asset code and scale (USD/2) instead of USD/9.

packages/wallet/backend/src/webMonetization/transaction/service.ts
```diff
@@ -57,20 +57,24 @@ export class WMTransactionService implements IWMTransactionService {
...
     const ids = transactions.map(({ id }) => id)
-    const sumResult = (await WMTransaction.query()
+    const sumResult = (await WMTransaction.query(trx)
       .whereIn('id', ids)
       .sum('value')) as unknown as [{ sum: bigint }]
 
     return {
       ids,
-      sum: sumResult[0].sum
+      sum: sumResult[0].sum ?? 0n
```
raducristianpopa, line 77: It returns `null` if there are no records.

packages/wallet/backend/src/app.ts
```diff
@@ -313,7 +313,29 @@ export class App {
...
+      .processWMPaymentPointers()
+      .catch((e) => {
+        logger.error(e)
+        return false
+      })
+      .then((trx) => {
+        if (trx) {
+          process.nextTick(() => this.processWMPaymentPointers())
+        } else {
+          setTimeout(() => this.processWMPaymentPointers(), 5000).unref()
```
raducristianpopa, line 332: 5 or 10 minutes for the WM worker?

packages/wallet/backend/src/paymentPointer/service.ts
dragosp1011:
```
      let webMonetizationAsset;
      if (args.isWM) {
        //* Web monetization feature requires an asset with scale MAX_ASSET_SCALE to exist. It's default assetCode is USD for now
        webMonetizationAsset =
            await this.deps.rafikiClient.getRafikiAsset(
                'USD',
                this.deps.env.MAX_ASSET_SCALE
            )

        if (!webMonetizationAsset) {
          throw new NotFound('Web monetization asset not found.')
        }
      }

      const assetId = webMonetizationAsset?.id || account.assetId
        const rafikiPaymentPointer =
          await this.deps.rafikiClient.createRafikiPaymentPointer(
            args.publicName,
              assetId,
            url
          )

        paymentPointer = await PaymentPointer.query().insert({
          url: rafikiPaymentPointer.url,
          publicName: args.publicName,
          accountId: args.accountId,
          id: rafikiPaymentPointer.id,
          isWM: args.isWM,
          assetCode: webMonetizationAsset?.code,
          assetScale: webMonetizationAsset?.scale
        })

      args.isWM &&
        (await this.deps.cache.set(paymentPointer.id, paymentPointer, {
          expiry: 60
        }))
```

packages/wallet/backend/src/app.ts
dragosp1011: both sound reasonable to me, lets leave it to 5

packages/wallet/backend/src/paymentPointer/service.ts
```diff
@@ -428,4 +428,166 @@ export class PaymentPointerService implements IPaymentPointerService {
...
+
+      throw new Error(
+        `Unable to transfer from ${source} into ${destination} error message: ${
+          transfer.status?.message || 'unknown'
+        }`
+      )
+    }
+
+    let updatePart: PartialModelObject<PaymentPointer> = {
```
dragosp1011, line 484: can be simplified a bit
```
const updatedField = type === 'OUTGOING' ? 'outgoingBalance' : 'incomingBalance'
const updatePart: PartialModelObject<PaymentPointer> = {
      [updatedField]: raw('?? - ?', [
        updatedField,
        this.deps.env.WM_THRESHOLD * balance
      ])
    }
```

packages/wallet/backend/src/rapyd/rapyd-client.ts
```diff
@@ -426,6 +426,7 @@ export class RapydClient implements IRapydClient {
       return res.data
     } catch (e) {
       if (e instanceof AxiosError) {
+        this.deps.logger.error(JSON.stringify(e.response?.data, null, 2))
```
beniaminmunteanu, line 429: so if e.response.data has content, it will be logged twice?

packages/wallet/backend/src/webMonetization/transaction/service.ts
```diff
@@ -53,24 +53,28 @@ export class WMTransactionService implements IWMTransactionService {
...
```
beniaminmunteanu, line 77: I know we don't enjoy doing calculations in JS, but since everything is bigint format, my take is that we'd end up fine with summing everything in JS and avoid another query.
But I'm fine with it this way as well

beniaminmunteanu (approved):
Looks fine

dragosp1011 (approved):

packages/wallet/frontend/src/pages/account/create.tsx
```diff
@@ -135,10 +135,12 @@ export const getServerSideProps: GetServerSideProps<{
...
-  const assets = response.data?.map((asset) => ({
-    value: asset.id,
-    label: asset.code
-  }))
+  const assets = response.data
+    ?.filter((asset) => asset.scale === 2)
+    ?.map((asset) => ({
+      value: asset.id,
+      label: asset.code
+    }))
```
raducristianpopa, line 143: To only display Rapyd supported assets and exclude the high scale ones.

dragosp1011: should it be less or equal with 2? (as in some cases in can be 0)

dragosp1011 (approved):

dragosp1011 (approved):

PR #936, opened 2023-10-26
feat(wallet/frontend): add cookie name environment variable

### LINKED ISSUE #935: Add `COOKIE_NAME` environment variable

raducristianpopa:
### Context

- fixes #935 

### Changes
To prevent cookie conflicts when deploying multiple instances under the same domain, we've introduced an environment variable called `COOKIE_NAME`. This variable helps the wallet frontend efficiently check the cookie in the middleware.


Tymmmy (approved):

dragosp1011 (approved):

docker/prod/docker-compose.yml
```diff
@@ -22,6 +22,7 @@ services:
       dockerfile: ./packages/wallet/frontend/Dockerfile.prod
       args:
         PORT: ${WALLET_FRONTEND_PORT}
+        COOKIE_NAME: ${WALLET_BACKEND_COOKIE_NAME}
```
dragosp1011, line 25: should the name be prefixed with `WALLET_FRONTEND`?

raducristianpopa: Used the same environment variable that we are using for the backend so we won't have two of them.

PR #938, opened 2023-10-26
feat(wallet): WM only for USD accounts

### LINKED ISSUE #937: Enable WM exclusively for USD

raducristianpopa:
### Context

- fixes #937

### DIFF
--- a/packages/boutique/backend/Dockerfile.dev
+++ b/packages/boutique/backend/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM node:lts-slim
+FROM node:hydrogen-slim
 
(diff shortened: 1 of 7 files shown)


packages/boutique/backend/Dockerfile.dev
```diff
@@ -1,4 +1,4 @@
-FROM node:lts-slim
+FROM node:hydrogen-slim
```
raducristianpopa, line 1: Node LTS is now at v20. We now have to specify what LTS version should be used.

dragosp1011 (approved):

Tymmmy (approved):

PR #942, opened 2023-10-27
feat: upgrade rafiki to alpha4

dragosp1011:
### Context

- fixes #


NirmitSeta:
I want to contribute here.

raducristianpopa (commented):
I find it weird to have "payment pointer" and "wallet address" at the same time ...

packages/boutique/backend/migrations/20231026113102_rename_payment_pointer_fields.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+  return Promise.any([
```
raducristianpopa, line 6:
```suggestion
  return Promise.all([
```

packages/wallet/backend/migrations/20231026113102_rename_payment_pointer_tables.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+  await knex.schema.renameTable('paymentPointers', 'walletAddresses')
+  return Promise.any([
```
raducristianpopa, line 7:
```suggestion
  return Promise.all([
```

Tymmmy:
@NirmitSeta this story is already in progress, and PR opened, and already some reviews here. This is close to being done. Thank you for your interest, but please check out stories from Interledger community where there are no assignees already. ANd then you can pick one and contribute.

Tymmmy:
@dragosp1011 I tried doing WIthdraw money for USD, and I get "message": "The request tried to create a payout, but a required field was missing or a field had an incorrect value. The request was rejected. Corrective action: Run 'Get Payout Required Fields' and check all values.",

packages/wallet/backend/src/rafiki/service.ts
```diff
@@ -21,14 +21,14 @@ export enum EventType {
   OutgoingPaymentCreated = 'outgoing_payment.created',
   OutgoingPaymentCompleted = 'outgoing_payment.completed',
   OutgoingPaymentFailed = 'outgoing_payment.failed',
-  PaymentPointerNotFound = 'payment_pointer.not_found'
+  WalletAddressNotFound = 'payment_pointer.not_found'
```
raducristianpopa, line 24:
```suggestion
  WalletAddressNotFound = 'wallet_address.not_found'
```

packages/wallet/backend/src/rafiki/rafiki-client.ts
```diff
@@ -212,6 +212,11 @@ export class RafikiClient implements IRafikiClient {
       if (response.withdrawEventLiquidity?.message === 'Transfer exists') {
         return true
       }
+
+      if (response.withdrawEventLiquidity?.message === 'Invalid id') {
+        this.deps.logger.debug(`Nothing to withdraw for event ${eventId}`)
+        return true
+      }
```
raducristianpopa, line 219: Instead of checking for the error message, we need to check the `balance` property that we receive on the webhook. If `balance` is 0, we do not need to call the `withdrawEventLiquidity` mutation.

raducristianpopa (approved):

PR #953, opened 2023-11-14
feat: Use Awilix as a DI container 

### LINKED ISSUE #811: Use Awilix as a DI container
Instead of using our own custom DI container, we should be move to a library that does this for us. The boutique's backend is already using a library called `awilix`.

https://github.com/interledger/testnet/blob/9fb0f475ddda91f0f23efa3b7e6df71a79951e38/packages/boutique/backend/src/container.ts#L42-L71

---
To do:
- [ ] Move wallet's backend from our custom DI container to `awilix`

rico191013:
### Context

- fixes #811


packages/wallet/backend/src/config/kenx.ts
raducristianpopa, line 1: Typo in filename.

raducristianpopa (commented):
I'm thinking about removing the `deps` object when doing DI and use the dependency directly (e.g.: `this.userService` instead of `this.deps.userService`). Thoughts on this?

packages/wallet/backend/src/config/kenx.ts
rico191013: Okay I'll resolve them all

dragosp1011 (dismissed):

raducristianpopa (dismissed):

raducristianpopa (approved):

dragosp1011 (approved):

Issue #683, opened 2023-09-01
Outgoing payment controller tests


Tymmmy:
Outgoing payment controller tests should be created.
Outgoing payment Service tests can be also added if needed.

ShatilKhan:
Is there any tutorial on how to add tests?
I'd like to work on this

PR #1017, opened 2023-12-13
test: added tests for outgoingPayment controller

rico191013:
### Context

- fixes https://github.com/interledger/testnet/issues/683

### Changes

Added tests for outgoing payment controller


packages/wallet/backend/tests/outgoingPayment/controller.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    await outgoingPaymentController.create(req, res, next)
+
+    expect(createByQuoteIdSpy).toHaveBeenCalled()
+    expect(createByQuoteIdSpy).toHaveBeenCalledTimes(1)
+    expect(createByQuoteIdSpy).toHaveBeenCalledWith(req.body.quoteId)
+  })
+
+  it('should return status 400 if the request body is invalid', async () => {
```
Tymmmy, line 67:
```suggestion
  it('should return status 400 if the request body is not valid', async () => {
```

packages/wallet/backend/tests/outgoingPayment/controller.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    )
+    createOutgoingPaymentControllerDepsMocked(false)
+  })
+
+  beforeEach(async (): Promise<void> => {
+    req = createRequest()
+    res = createResponse()
+  })
+
+  it('should call createByQuoteId in outgoingPaymentService.', async () => {
```
Tymmmy, line 53:
```suggestion
  it('should call createByQuoteId in outgoingPaymentService', async () => {
```

Tymmmy: I mean, we should decide if we put everywhere '.' or nowhere I think

Tymmmy (dismissed):

Tymmmy (approved):

PR #1018, opened 2023-12-14
test: Add Rafiki Controller Tests

### LINKED ISSUE #686: Rafiki controller tests
Rafiki controller tests should be created.
Rafiki Service tests can be also added if needed.

rico191013:
closes #686 

### Changes

We Added some Tests for Rafiki controller, we tried to make it very similar to old tests we found on the project, so if there is something wrong we will be able to change it and make it more close to the projects structure.

@aosccode


packages/wallet/backend/tests/rafiki/controller.test.ts
```diff
@@ -0,0 +1,170 @@
...
+      req.body = mockOnWebhookRequest().body
+      delete req.body.type
+      const onWebHookSpy = jest.spyOn(mockRafikiService, 'onWebHook')
+
+      await rafikiController.onWebHook(req, res, next)
+
+      expect(onWebHookSpy).not.toBeCalled()
+    })
+
+    it('should call logger error if the request body is invalid', async () => {
```
Tymmmy, line 159:
```suggestion
    it('should call logger error if the request body is not valid', async () => {
```

Tymmmy (dismissed):

Tymmmy (approved):

Issue #682, opened 2023-09-01
Incoming payment Controller Tests


Tymmmy:
Incoming payment controller tests should be created

jocaetano1:
Hi, I am interest to resolve this issue.

PR #1019, opened 2023-12-14
test: Add tests in Incoming Payment Controller 

rico191013:
### Context

- closes (or fixes) #682

Test in **IncomingPaymentController**

### Changes

We've added tests to the IncomingPaymentController.

You can check the changes.

@aosccode


packages/wallet/backend/tests/rafiki/controller.test.ts
```diff
@@ -0,0 +1,170 @@
...
+      const resSpy = jest.spyOn(res, 'send')
+
+      await rafikiController.onWebHook(req, res, next)
+
+      expect(res.statusCode).toBe(200)
+      expect(resSpy).toBeCalled()
+      expect(resSpy).toBeCalledTimes(1)
+    })
+
+    it('should return status 400 if the request body is invalid', async () => {
```
Tymmmy, line 135:
```suggestion
    it('should return status 400 if the request body is not valid', async () => {
```

packages/wallet/backend/tests/rafiki/controller.test.ts
```diff
@@ -0,0 +1,170 @@
...
+        next()
+        errorHandler(err, req, res, next)
+      })
+
+      expect(next).toBeCalled()
+      expect(next).toBeCalledTimes(1)
+      expect(res.statusCode).toBe(400)
+    })
+
+    it('should not call onWebHook in rafikiService if the request body is invalid', async () => {
```
Tymmmy, line 149:
```suggestion
    it('should not call onWebHook in rafikiService if the request body is not valid', async () => {
```

packages/wallet/backend/tests/rafiki/controller.test.ts
```diff
@@ -0,0 +1,170 @@
...
+      req.body = mockOnWebhookRequest().body
+      delete req.body.type
+      const onWebHookSpy = jest.spyOn(mockRafikiService, 'onWebHook')
+
+      await rafikiController.onWebHook(req, res, next)
+
+      expect(onWebHookSpy).not.toBeCalled()
+    })
+
+    it('should call logger error if the request body is invalid', async () => {
```
Tymmmy, line 159:
```suggestion
    it('should call logger error if the request body is not valid', async () => {
```

Tymmmy (approved):

PR #1033, opened 2023-12-28
feat(grants): show interaction only for PENDING grants

### LINKED ISSUE #721: Grant interaction improovement
A link to accept or deny grants can be opened from the user many times, and so the Grant can be accepted, declined, accepted, declined etc...multiple times.
We should limit the user to access the link only once, if the Grant was already accepted or declined the user should receive a message that the Grant was already handled and redirected to Grants list.

dragosp1011:
### Context

- fixes #721


Tymmmy:
Tested the grant interaction for re-interaction. Works ok.

Tymmmy (approved):

PR #1069, opened 2024-01-12
feat: Rpyd to zod schema

### LINKED ISSUE #378: Rapyd return types as zod schemas

rico191013:
### Context

- fixes #378


beniaminmunteanu (approved):
I'm approving with the note that this was previously in a [branch](https://github.com/interledger/testnet/compare/main...rapyd-zod-validation) that had a draft PR opened for a long time, as it was not urgent and I was waiting for alternative solutions.
Since i had my go on it, and Rico did as well, I'm approving it.

Maybe @dragosp1011  should have a final look as well, and if we don't have another say in the matter, just merge this.

dragosp1011 (changes requested):
The whole idea of adding zod schemas for rapyd responses is to validate that the response for them is the one that we expect.
Changing the interfaces into zod schemas does not help us in any way.

@beniaminmunteanu work on a previous [PR](https://github.com/interledger/testnet/pull/440/files) was pretty solid, check his PR and see maybe you have any idea how to solve in a better way the issues that Beni was facing there.

`.d.ts` files are used for interfaces and types, the defined schemas should go in a separate file.

beniaminmunteanu:
Oh i see that the schemas are not actually used for validations, but just for types inference. 
Didn't notice it @dragosp1011, good catch.

I would also suggest Rico considers just updating that PR that i had open to the current main, or I can do it next week as well.
If there is no other alternative solution to mine on array-like rapyd response, then that PR just needs to have main merged into it or a rebase, and maybe if other rapyd calls were added/changed in the meanwhile, update those.

PR #1101, opened 2024-01-27
feat: Seed wallet backend with default account

### LINKED ISSUE #749: Seed wallet with default accounts and payments pointer
We want to make installing local environment very easy for new developers.
For that we want to already have Testnet Boutique up and running on local environment with simply installing the project locally.

What we need:
Backend: 
- when deploying the project locally, there should be a background creation of 2 users

1. one user 'dev@email.com' , where all checks for email are done for this fake email, create user in Rapyd if it is the first time, or all Rapyd values from KYC reused/reseeded into DB, if the user 'dev@email.com' was already created in RAPYD, and rapyd calls skipped. Password can be '123456'.
2. the second user should be for the BOUTIQUE, user 'boutique@mail.com' (EDIT - let's use 'boutique@email.com')
, the same as the above user, all checks are done in the backgorund to accept this fake email, create user in RAPYD if it is the first time, if not, reuse Rapyd KYC data, so that we don't get error that the email was already used. Password can be '123456'.

- replace the default EUR account and payment pointer creation with a USD account and USD payment pointer
-  for user 'dev@email.com' payment pointer name can be the one generated randomly
- the generated USD payment pointer for user 'boutique@email.com', should be named 'boutique'
- we need to generate Developer Keys for the 'boutique' payment pointer, and reuse them all the time, so that the PRIVATE_KEY and KEY_ID values from env file can be saved in docker compose file and always used from there, with the same value
- update .env.example and docker-compose file with the changes
- test if everything is working after first deploy of testnet, then clean docker and redeploy and test again

Frontend:

- Replace Onboarding USD with Onboarding EUR
- Update Readme file with the necesarry changes

rico191013:
### Context

- fixes #749

### DIFF
--- a/packages/wallet/backend/src/account/service.ts
+++ b/packages/wallet/backend/src/account/service.ts
@@ -291,10 +291,11 @@ export class AccountService implements IAccountService {
   }
 
   public async createDefaultAccount(
-    userId: string
+    userId: string,
+    assetCode = 'EUR'
   ): Promise<Account | undefined> {
     const asset = (await this.rafikiClient.listAssets({ first: 100 })).find(
-      (asset) => asset.code === 'EUR' && asset.scale === 2
+      (asset) => asset.code === assetCode && asset.scale === 2
     )
     if (!asset) {
       return
--- a/packages/wallet/backend/src/auth/service.ts
+++ b/packages/wallet/backend/src/auth/service.ts
@@ -55,6 +55,9 @@ export class AuthService implements IAuthService {
   }
 
   public async authorize(args: AuthorizeArgs): Promise<AuthorizeResult> {
+    if (args.email === this.env.DEFAULT_AUTH_USERNAME)
+      await this.userService.createDefaultAccount()
+
     const user = await this.userService.getByEmail(args.email)
 
     // TODO: Prevent timing attacks
--- a/packages/wallet/backend/src/config/env.ts
+++ b/packages/wallet/backend/src/config/env.ts
@@ -35,7 +35,13 @@ const envSchema = z.object({
   BASE_ASSET_SCALE: z.coerce.number().nonnegative().default(2),
   MAX_ASSET_SCALE: z.coerce.number().nonnegative().default(9),
   WM_THRESHOLD: z.coerce.bigint().nonnegative().default(100_000_000n), // $0.1 in asset scale 9
-  DEBT_THRESHOLD: z.coerce.number().multipleOf(0.01).nonnegative().default(5.0) // $5.00
+  DEBT_THRESHOLD: z.coerce.number().multipleOf(0.01).nonnegative().default(5.0), // $5.00
+  DEFAULT_AUTH_USERNAME: z.string().default('dev@email.com'),
+  DEFAULT_AUTH_PASSWORD: z.string().default('123456'),
+  DEFAULT_WALLET_ID: z
+    .string()
+    .default('ewallet_041c78fb4629be9cfc6387585cf768d5'),
+  DEFAULT_KYC_ID: z.string().default('kycid_f28831c9d3f573bd87daa773e9c1cc03')
 })
 
 export type Env = z.infer<typeof envSchema>
--- a/packages/wallet/backend/src/user/service.ts
+++ b/packages/wallet/backend/src/user/service.ts
@@ -3,6 +3,7 @@ import { User } from './model'
 import { EmailService } from '@/email/service'
 import { getRandomToken, hashToken } from '@/utils/helpers'
 import { Logger } from 'winston'
+import { Env } from '@/config/env'
 
 interface CreateUserArgs {
   email: string
@@ -22,7 +23,8 @@ interface IUserService {
 export class UserService implements IUserService {
   constructor(
     private emailService: EmailService,
-    private logger: Logger
+    private logger: Logger,
+    private env: Env
   ) {}
 
   public async create(args: CreateUserArgs): Promise<User> {
@@ -110,6 +112,19 @@ export class UserService implements IUserService {
     return !!user
   }
 
+  public async createDefaultAccount() {
+    const existingUser = await this.getByEmail(this.env.DEFAULT_AUTH_USERNAME)
+
+    if (existingUser) return
+
+    const args = {
+      email: this.env.DEFAULT_AUTH_USERNAME,
+      password: this.env.DEFAULT_AUTH_PASSWORD,
+      isEmailVerified: true
+    }
+
+    return User.query().insertAndFetch(args)
+  }
   public async verifyEmail(token: string): Promise<void> {
     const verifyEmailToken = hashToken(token)
 
--- a/packages/wallet/backend/src/rapyd/controller.ts
+++ b/packages/wallet/backend/src/rapyd/controller.ts
@@ -11,6 +11,7 @@ import {
   RapydIdentityResponse,
   RapydWallet
 } from './schemas'
+import { Env } from '@/config/env'
 
 interface IRapydController {
   getCountryNames: ControllerFunction<Options[]>
@@ -24,7 +25,8 @@ export class RapydController implements IRapydController {
   constructor(
     private accountService: AccountService,
     private walletAddressService: WalletAddressService,
-    private rapydService: RapydService
+    private rapydService: RapydService,
+    private env: Env
   ) {}
 
   public getCountryNames = async (
@@ -74,22 +76,29 @@ export class RapydController implements IRapydController {
         body: { firstName, lastName, address, city, country, zip, phone }
       } = await validate(walletSchema, req)
 
-      const createWalletResponse = await this.rapydService.createWallet({
-        firstName,
-        lastName,
-        address,
-        city,
-        country,
-        zip,
-        email,
-        id,
-        phone
-      })
+      const isDefault = email === this.env.DEFAULT_AUTH_USERNAME
+      const createWalletResponse = await this.rapydService.createWallet(
+        {
+          firstName,
+          lastName,
+          address,
+          city,
+          country,
+          zip,
+          email,
+          id,
+          phone
+        },
+        isDefault
+      )
 
       req.session.user.needsWallet = false
       await req.session.save()
 
-      const defaultAccount = await this.accountService.createDefaultAccount(id)
+      const defaultAccount = await this.accountService.createDefaultAccount(
+        id,
+        isDefault ? 'USD' : 'EUR'
+      )
       if (defaultAccount) {
         const typedArray = new Uint32Array(1)
         getRandomValues(typedArray)
--- a/packages/wallet/backend/src/rapyd/rapyd-client.ts
+++ b/packages/wallet/backend/src/rapyd/rapyd-client.ts
@@ -131,6 +131,10 @@ export class RapydClient implements IRapydClient {
     return this.put('user', JSON.stringify(profile), RapydWalletSchema)
   }
 
+  public getProfile(walletID: string): Promise<RapydResponse<RapydWallet>> {
+    return this.get(`user/${walletID}`, RapydWalletSchema)
+  }
+
   public verifyIdentity(
     req: RapydIdentityRequest
   ): Promise<RapydResponse<RapydIdentityResponse>> {
--- a/packages/wallet/backend/src/rapyd/schemas.ts
+++ b/packages/wallet/backend/src/rapyd/schemas.ts
@@ -184,8 +184,8 @@ export const RapydAccountSchema = z.object({
   alias: z.string().optional(),
   balance: z.number().optional(),
   currency: z.string().optional(),
-  limit: z.number().optional(),
-  limits: z.number().optional(),
+  limit: z.number().optional().nullable(),
+  limits: z.number().optional().nullable(),
   on_hold_balance: z.number().optional(),
   received_balance: z.number().optional(),
   reserve_balance: z.number().optional()
--- a/packages/wallet/backend/src/rapyd/service.ts
+++ b/packages/wallet/backend/src/rapyd/service.ts
@@ -8,6 +8,7 @@ import {
   RapydIdentityResponse,
   RapydWallet
 } from './schemas'
+import { Env } from '@/config/env'
 
 type VerifyIdentityParams = {
   userId: string
@@ -54,7 +55,10 @@ export interface IRapydService {
 }
 
 export class RapydService implements IRapydService {
-  constructor(private rapydClient: RapydClient) {}
+  constructor(
+    private rapydClient: RapydClient,
+    private env: Env
+  ) {}
 
   public async getDocumentTypes(userId: string) {
     const user = await User.query().findById(userId)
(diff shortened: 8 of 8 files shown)


docker/dev/docker-compose.yml
```diff
@@ -70,9 +70,9 @@ services:
       NODE_ENV: development
       PORT: 3004
       DATABASE_URL: postgres://boutique_backend:boutique_backend@postgres/boutique_backend
-      PRIVATE_KEY: ${PRIVATE_KEY}
-      KEY_ID: ${KEY_ID}
-      PAYMENT_POINTER: ${PAYMENT_POINTER}
+      PRIVATE_KEY: LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1DNENBUUF3QlFZREsyVndCQ0lFSU9JMEtXWktpVE1uU2FnWDF5K1l1bW12dWY2S2d4QThPNm1JM2ZMRm9xVzcKLS0tLS1FTkQgUFJJVkFURSBLRVktLS0tLQ==
+      KEY_ID: d0c7b0a0-bc95-4b96-907e-7a0e3c79d80e
+      PAYMENT_POINTER: $rafiki-backend/d839fdd7
```
Tymmmy, line 75: Why the generated name?
We would like to have a 'boutique' name here, to be always used

As the story says:
'the generated USD payment pointer for user '[boutique@email.com](mailto:boutique@email.com)', should be named 'boutique''

packages/wallet/backend/src/account/service.ts
```diff
@@ -291,16 +296,26 @@ export class AccountService implements IAccountService {
   }
 
   public async createDefaultAccount(
-    userId: string
+    userId: string,
+    type: DefaultAccountType = DefaultAccountType.EUR
   ): Promise<Account | undefined> {
+    let assetCode = 'EUR'
```
Tymmmy, line 302: 'replace the default EUR account and payment pointer creation with a USD account and USD payment pointer' -> from story, I understand here, that we can remove EUR default account and PP creation, and just leave in USD creation

packages/wallet/backend/src/account/service.ts
```diff
@@ -291,16 +296,26 @@ export class AccountService implements IAccountService {
...
+      (asset) => asset.code === assetCode && asset.scale === 2
     )
     if (!asset) {
       return
     }
     const account = await this.createAccount({
-      name: 'EUR Account',
+      name:
+        type === DefaultAccountType.USD_BOUTIQUE
+          ? 'boutique'
```
Tymmmy, line 317: The account name can be 'boutique' but then let's do 'Boutique'

packages/wallet/backend/src/config/env.ts
```diff
@@ -35,7 +35,16 @@ const envSchema = z.object({
   BASE_ASSET_SCALE: z.coerce.number().nonnegative().default(2),
   MAX_ASSET_SCALE: z.coerce.number().nonnegative().default(9),
   WM_THRESHOLD: z.coerce.bigint().nonnegative().default(100_000_000n), // $0.1 in asset scale 9
-  DEBT_THRESHOLD: z.coerce.number().multipleOf(0.01).nonnegative().default(5.0) // $5.00
+  DEBT_THRESHOLD: z.coerce.number().multipleOf(0.01).nonnegative().default(5.0), // $5.00
+  DEFAULT_AUTH_USERNAME: z.string().default('dev@email.com'),
+  DEFAULT_BOUTIQUE_AUTH_USERNAME: z.string().default('boutique@mail.com'),
```
Tymmmy, line 40: Sorry for the typo in the story, let's do this also 'email' and not 'mail'

Tymmmy:
- I tried to go directly to localhost/boutique and shop around. I had an 'Invalid payment pointer' error. Details in docker boutique-backend: 'Could not fetch payment pointer "-backend/d839fdd7".' This means that the Private key and key id is not associated with boutique PP correctly. We somehow need to seed the values from docker.compose to the DB as well, so boutique can use them?
- If the boutique PP already has Developer keys generated and correctly seeded into DB, then also the Settings page Developer keys for the PP should be updated accordingly. Now I have empty keys for the PP (should be populated with values):
image

packages/wallet/backend/src/account/service.ts
```diff
@@ -291,16 +296,26 @@ export class AccountService implements IAccountService {
...
+          : `${type} Account`,
```
Tymmmy, line 318: it creates 'usd Account' -> We would like it to be USD Account

Tymmmy:
I saw that both accounts created, Dev & Boutique have funds in their USD accounts, 500$:
image

When we go to the Transactions page we see that there was a Fund account action done, but with 100$. Why is that difference?
image

packages/wallet/backend/src/user/service.ts
```diff
@@ -110,6 +117,53 @@ export class UserService implements IUserService {
...
+        walletAddressName,
+        publicName: 'Default Payment Pointer',
+        userId: createdUser.id,
+        isWM: false
+      })
+      await this.walletAddressService.registerKey(
+        createdUser.id,
+        defaultAccount.id,
+        wallet.id,
+        {
```
Tymmmy, line 157: keys should be registered only for the 'boutique' PP, right now we have the Developer accounts USD payment pointer having the same keys as the boutique, which is probably an error. So only save keys for 'boutique' PP

Tymmmy:
Please also check after new deploy, we should be able to directly shop around in the boutique. We are gettings error at checkout:
image

With docker error:
2024-02-06 10:54:16 [2024-02-06 08:54:16] [error] Could not fetch payment pointer "-backend/boutique".

This means I guess 'boutique' PP is not recognized as a correct PP that is associated with Boutique, this should be the PP that receives the money after users buy. This is the cashier of boutique, and it is not working. Is it getting correct Developer keys for it?

docker/dev/docker-compose.yml
```diff
@@ -70,9 +70,9 @@ services:
       NODE_ENV: development
       PORT: 3004
       DATABASE_URL: postgres://boutique_backend:boutique_backend@postgres/boutique_backend
-      PRIVATE_KEY: ${PRIVATE_KEY}
-      KEY_ID: ${KEY_ID}
-      PAYMENT_POINTER: ${PAYMENT_POINTER}
+      PRIVATE_KEY: LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1DNENBUUF3QlFZREsyVndCQ0lFSUI0YzgyOVlSMDZCTUhtQmpIVTNrOHZmMEZhVEFvOHNCTzIrRFZwY1lPSk0KLS0tLS1FTkQgUFJJVkFURSBLRVktLS0tLQ==
```
Tymmmy, line 73: Boutiques is not working, so I am guessing a key is not correct. Looking back on previous implementation locally with using keys from environment variables, I have the Private Key with only one '=' at the very end. You have '=='.
I think this is the new line added into encoding. Please try with one '='.

Tymmmy:
> > I saw that both accounts created, Dev & Boutique have funds in their USD accounts, 500$: image
> > When we go to the Transactions page we see that there was a Fund account action done, but with 100$. Why is that difference? image
> > EDIT - as many times as you redeploy localy, there is a sum of 100$ added to the previous amount. So now I am on my 6th redeploy and instead of having just 100$ in my account, I have 600$. Can we fix this? :)
> 
> Well, it's because of the same wallet account for the Rapyd platform. Since we use a specific wallet for Rapyd and do not create a new one for each deployment, it causes this difference.

Can we fix it somehow? To not have this inconsistency? Withdraw the money from the account before creating again?

rico191013:
> inconsistency

Fixed, just increased the timeout amount because it received a timeout error for the first time when the user was in the creation process.

docker/dev/docker-compose.yml
```diff
@@ -77,9 +77,8 @@ services:
...
```
dragosp1011, line 80: IMO the default values should be in `.env.example` and not in docker-compose. That will allow devs to update them as needed without affecting the code

beniaminmunteanu: I agree.

Tymmmy:
1. We get Invalid payment pointer error in checkout after shopping with the dev@email accounts PP if we didn't log in with boutique user beforehand. This is not ok, we shouldn't make obligatory logging in to boutique@email.com user before we go shopping. I guess you are calling the functions from backend to create boutique user and settings for it, but I am sure this is what Radu meant in the original story to SEED WALLET-> probably SEED DB wth values correctly, instead of calling backend functions to insert into DB. 
Docker logs error -> fo] received webhook of type : wallet_address.not_found for : 
2024-02-14 13:23:29 [2024-02-14 11:23:29] [warn] wallet_address.not_found received
2. After I log in to boutique user, checkout give SUCCESS, everything seems ok, until you go to transactions list, and see that payment is PENDING for a long time. Checking docker logs we see error:
Webhook response error for rafiki: Request failed with status code 400
2024-02-14 13:25:13 [2024-02-14 11:25:13] [debug] {
2024-02-14 13:25:13   "status": {
2024-02-14 13:25:13     "error_code": "WALLET_TRANSFER_NOT_ALLOWED_FOR_COLLECT_WALLET",
2024-02-14 13:25:13     "status": "ERROR",
2024-02-14 13:25:13     "message": "The request tried to transfer funds from a wallet to wallet, but one of the wallets was a client wallet of type 'collect' and this action is not allowed. The request was rejected. Corrective action: None.",
2024-02-14 13:25:13     "response_code": "WALLET_TRANSFER_NOT_ALLOWED_FOR_COLLECT_WALLET",
2024-02-14 13:25:13     "operation_id": "f7fa8004-dfc8-46a0-a21b-e1cd1ef52473"
2024-02-14 13:25:13   }
2024-02-14 13:25:13 }

3. Very very confusing for the dev user: 
image

If we can't make it that we always have 100$ funds for the 2 users, and we always need to withdraw funds in order to always have 100$, then let's go back to not withdrawing funds, and the user having just more money. This way they just have more money, not a confusing withdraw from account.

4. Creating a new user gives me error when I do KYC:
image

packages/wallet/backend/src/user/service.ts
```diff
@@ -110,6 +117,66 @@ export class UserService implements IUserService {
     return !!user
   }
 
+  public async createDefaultAccount(email: string) {
+    const existingUser = await this.getByEmail(email)
+
+    if (existingUser) return
+
+    let defaultUser = this.env.DEFAULT_WALLET_ACCOUNT
```
dragosp1011, line 125: `const defaultUser = email === this.env.DEFAULT_BOUTIQUE_ACCOUNT.email ? this.env.DEFAULT_BOUTIQUE_ACCOUNT : this.env.DEFAULT_WALLET_ACCOUNT`

packages/wallet/backend/src/walletAddress/service.ts
```diff
@@ -292,21 +297,38 @@ export class WalletAddressService implements IWalletAddressService {
   async registerKey(
     userId: string,
     accountId: string,
-    walletAddressId: string
+    walletAddressId: string,
+    defaultAccount:
+      | { publicKeyPEM: string; privateKeyPEM: string; keyId: string }
+      | undefined = undefined
```
dragosp1011, line 303: make the param optional instead

docker/dev/docker-compose.yml
raducristianpopa: Replying to @dragosp1011 and @beniaminmunteanu: I agree, the values should be in `.env.example`.

As a side note: I think that seeding process for the initial setup might be better handled as a script. This script could either run automatically at startup or be executed manually, rather than being integrated directly into the backend.

packages/wallet/backend/src/auth/service.ts
```diff
@@ -55,6 +55,12 @@ export class AuthService implements IAuthService {
   }
 
   public async authorize(args: AuthorizeArgs): Promise<AuthorizeResult> {
+    if (
+      args.email === this.env.DEFAULT_WALLET_ACCOUNT.email ||
+      args.email === this.env.DEFAULT_BOUTIQUE_ACCOUNT.email
+    )
+      await this.userService.createDefaultAccount()
```
dragosp1011, line 62: dafault users should be created at startup or using as script as @raducristianpopa suggested

with current approach if I register a new account and login there, I won't be able to use the boutique as the default account for it will still not be created

packages/wallet/backend/src/user/service.ts
```diff
@@ -109,7 +118,76 @@ export class UserService implements IUserService {
...
+      false
+    )
+
+    args = {
+      ...defaultBoutiqueUser,
+      isEmailVerified: true
+    }
+    const createdBoutiqueUser = await User.query().insertAndFetch(args)
+
+    const boutiquePromise = this.accountService.createDefaultAccount(
```
dragosp1011, line 150: create user + default account are similar and can be extracted in one method that will be called with default user details

packages/wallet/backend/src/user/service.ts
```diff
@@ -109,7 +118,76 @@ export class UserService implements IUserService {
...
     return !!user
   }
+  public async createDefaultAccount() {
+    const existingUser = await this.getByEmail(
+      this.env.DEFAULT_WALLET_ACCOUNT.email
+    )
 
+    if (existingUser) return
+
+    await this.rafikiClient.createAsset('USD', 2)
```
dragosp1011, line 128: what will happen if the asset already exists?

packages/wallet/backend/src/app.ts
```diff
@@ -344,4 +344,9 @@ export class App {
     process.nextTick(() => this.processPendingTransactions())
     process.nextTick(() => this.processWMWalletAddresses())
   }
+
+  async createDefaultUsers() {
+    const userService = this.container.resolve('userService')
+    await userService.createDefaultAccount()
```
dragosp1011, line 350: will be useful to have a message when the default accounts are created

Tymmmy: @dragosp1011 are you thinking a message for the user? I was thinking of updating README and explaining there for local dev people. Would that be enough?

dragosp1011: a log in backend app to say that default accounts are ready will be helpful to know when they can be used

packages/wallet/backend/src/index.ts
```diff
@@ -16,6 +16,7 @@ async function bootstrap() {
     console.log(e)
   })
   await app.processResources()
+  await app.createDefaultUsers()
```
dragosp1011, line 19: this should be called only in dev environment

packages/wallet/backend/src/user/service.ts
```diff
@@ -109,7 +118,77 @@ export class UserService implements IUserService {
...
     return !!user
   }
+  public async createDefaultAccount() {
+    const existingUser = await this.getByEmail(
+      this.env.DEFAULT_WALLET_ACCOUNT.email
+    )
+
+    if (existingUser) return
+
+    const asset = (await this.rafikiClient.listAssets({ first: 100 })).find(
```
dragosp1011, line 128: use `getRafikiAsset` method instead

packages/wallet/backend/src/user/service.ts
```diff
@@ -109,6 +118,91 @@ export class UserService implements IUserService {
...
+          privateKeyPEM: this.env.DEFAULT_BOUTIQUE_KEYS.private_key,
+          keyId: this.env.DEFAULT_BOUTIQUE_KEYS.key_id
+        }
+      )
+    }
+
+    this.logger.info('Default users have been successfully created')
+  }
+
+  private async createBoutiqueDefaultUser(
```
dragosp1011, line 173: both methods do the same thing.
as below only the user id is used form user info and `userId` exists in account details, returning the new account should be enough

packages/wallet/backend/src/account/service.ts
```diff
@@ -300,16 +300,17 @@ export class AccountService implements IAccountService {
   }
 
   public async createDefaultAccount(
-    userId: string
+    userId: string,
+    isBoutique: boolean = false
```
dragosp1011, line 304: change `isBoutique` param to `name` with default value 'USD Account'
this why the method can  be used in the future for other type of accounts without any changes

packages/wallet/backend/src/walletAddressKeys/service.ts
```diff
@@ -93,21 +98,37 @@ export class WalletAddressKeyService implements IWalletAddressKeyService {
     userId,
     accountId,
     walletAddressId,
-    nickname
+    nickname,
+    defaultAccount
```
dragosp1011, line 102: rename to `keyPair`

dragosp1011 (approved):

PR #1106, opened 2024-02-01
Grant list improvements

### LINKED ISSUE #757: Grant list and details improvements
EDIT - implement the new API call for grant list, and show pagination.
Check what other user friendly improvements can be made.

For the Grants list we should order the list from latest to oldest, also adding pagination for the list.
The Grants list should display Public name of client PP, public name should also be available in the details page.
Currently UI only receives the client PP and that is displayed.

After the public name is displayed for Grants, add a new column with the status of the grant.

Tymmmy:
### Context

- fixes #757 

### Changes
- Added pagination to grants list
- Added more details to the list page
- Added filters regarding status

### DIFF
--- a/packages/wallet/frontend/src/lib/api/grants.ts
+++ b/packages/wallet/frontend/src/lib/api/grants.ts
@@ -1,3 +1,4 @@
+import { z } from 'zod'
 import {
   getError,
   httpClient,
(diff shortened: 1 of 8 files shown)


packages/wallet/frontend/src/lib/api/grants.ts
```diff
@@ -57,7 +87,8 @@ type FinalizeInteractionParams = {
 type FinalizeInteractionResponse = SuccessResponse | ErrorResponse
 
 interface GrantsService {
-  list: (cookies?: string) => Promise<ListGrantsResponse>
+  list: (args?: GrantListArgs, cookies?: string) => Promise<GrantsListResponse>
```
rico191013, line 90: Maybe it could be good to rename list -> filter, listAll -> list?

dragosp1011 (approved):

packages/wallet/frontend/src/lib/api/grants.ts
Tymmmy: Well...it's not really filtering though, it is listing all in a specific pagination setup

PR #1122, opened 2024-02-09
feat: allow multiple keys per wallet address

dragosp1011:
### Context

- fixes #

### Changes
- moved wallet address keys in a separate table
- added endpoint to upload JWK
- added endpoint to update key name
- added endpoint to list keys for an wallet addres

### DIFF
--- a/packages/wallet/backend/src/walletAddressKeys/controller.ts
+++ b/packages/wallet/backend/src/walletAddressKeys/controller.ts
@@ -1,0 +1,154 @@
+import { validate } from '@/shared/validate'
+import type { NextFunction, Request } from 'express'
+import { KeyResponse, WalletAddressKeyService } from './service'
+import {
+  generateWalletAddressKey,
+  patchWalletAddressKey,
+  uploadWalletAddressKey
+} from './validation'
+import { WalletAddressKeys } from '@/walletAddressKeys/model'
+
+interface IWalletAddressKeyController {
+  registerKey: ControllerFunction<KeyResponse>
+  revokeKey: ControllerFunction
+}
+
+interface KeyPair {
+  publicKey: string
+  privateKey: string
+  keyId: string
+}
+
+export class WalletAddressKeyController implements IWalletAddressKeyController {
+  constructor(private walletAddressKeyService: WalletAddressKeyService) {}
+
+  list = async (
+    req: Request,
+    res: CustomResponse<WalletAddressKeys[]>,
+    next: NextFunction
+  ) => {
+    try {
+      const userId = req.session.user.id
+      const { accountId, walletAddressId } = req.params
+      const keys = await this.walletAddressKeyService.lisByWalletId({
+        userId,
+        accountId,
+        walletAddressId
+      })
+
+      res.status(200).json({
+        success: true,
+        message: 'SUCCESS',
+        result: keys
+      })
+    } catch (e) {
+      next(e)
+    }
+  }
+
+  patchKey = async (
+    req: Request,
+    res: CustomResponse<KeyPair>,
+    next: NextFunction
+  ) => {
+    try {
+      const userId = req.session.user.id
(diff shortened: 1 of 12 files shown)


packages/wallet/backend/src/walletAddressKeys/controller.ts
```diff
@@ -0,0 +1,154 @@
...
+        userId,
+        accountId,
+        walletAddressId,
+        base64Key,
+        name
+      })
+
+      res.status(200).json({
+        success: true,
+        message: 'Public key is successfully uploaded'
```
Tymmmy, line 128:
```suggestion
        message: 'Public key is successfully uploaded.'
```

packages/wallet/backend/src/walletAddressKeys/controller.ts
```diff
@@ -0,0 +1,154 @@
...
+      const keyResponse = await this.walletAddressKeyService.registerKey({
+        userId,
+        accountId,
+        walletAddressId,
+        name
+      })
+
+      res.status(200).json({
+        success: true,
+        message: 'Public key is successfully registered',
```
Tymmmy, line 98:
```suggestion
        message: 'Public key is successfully registered.',
```

packages/wallet/backend/src/walletAddressKeys/controller.ts
```diff
@@ -0,0 +1,154 @@
...
+      const {
+        body: { name, base64Key }
+      } = await validate(uploadWalletAddressKey, req)
+
+      await this.walletAddressKeyService.uploadKey({
+        userId,
+        accountId,
+        walletAddressId,
+        base64Key,
+        name
```
Tymmmy, line 123: Let's use nickname

packages/wallet/backend/src/walletAddressKeys/service.ts
```diff
@@ -0,0 +1,206 @@
...
+  keyId: string
+  nickname: string
+}
+
+interface IWalletAddressKeyService {
+  registerKey: (params: RegisterKeyArgs) => Promise<KeyResponse>
+  revokeKey: (params: RevokeKeyArgs) => Promise<void>
+  uploadKey: (params: UploadKeyArgs) => Promise<void>
+  patch: (params: PatchKeyArgs) => Promise<void>
+  lisByWalletId: (params: WalletAddressKeyArgs) => Promise<WalletAddressKeys[]>
```
Tymmmy, line 45:
```suggestion
  listByWalletId: (params: WalletAddressKeyArgs) => Promise<WalletAddressKeys[]>
```

packages/wallet/backend/src/walletAddressKeys/service.ts
```diff
@@ -0,0 +1,206 @@
...
+      walletAddressId
+    })
+
+    const jwk = validateJwk(
+      JSON.parse(new Buffer(base64Key, 'base64').toString())
+    )
+    const publicKey = createPublicKey({ key: jwk, format: 'jwk' })
+    const publicKeyPEM = publicKey
+      .export({ type: 'spki', format: 'pem' })
+      .toString()
```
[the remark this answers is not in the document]
dragosp1011: as this was the format that we kept the public key before, kept it for the uploaded keys as well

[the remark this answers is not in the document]
dragosp1011: also regarding this, do we need to store the public key at all in our db?
does not look like we do anything with it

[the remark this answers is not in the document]
raducristianpopa: At this moment, we allow the users to visualize the public key in the Developer Keys section.

beniaminmunteanu (approved):
LGTM. 
Only suggestion i would have is that the getAccounts method receives both include params as an object. 
But since it's just 2 optional "includes", I'm fine with it like this as well.

PR #1126, opened 2024-02-13
feat: Add soft delete for transactions

### LINKED ISSUE #312: Soft delete functionality improvements
Context: https://github.com/interledger/testnet/pull/292#discussion_r1176325628

User is able to delete the payment pointer, but right now there is a soft delete happening, basically in the DB the PP becomes inactive. When creating a PP with the same name as the deleted one, the previous transactions will be assigned to the new one. This is not desired, so we will add a flag for transactions, that they are old transactions-deprecated, and we will not display them anymore.
We need to improve on the PP delete functionality.
 When the user creates a new PP but with the same name as the one previously deleted, then the transactions done in the past will not appear anymore.

rico191013:
### Context

- fixes #312


packages/wallet/backend/migrations/20230728101910_update_transactions_columns.js
```diff
@@ -8,7 +8,7 @@ exports.up = function (knex) {
     await knex.raw(
       [
         `ALTER TABLE "transactions" DROP CONSTRAINT IF EXISTS "transactions_status_check";`,
-        `ALTER TABLE "transactions" ADD CONSTRAINT "transactions_status_check" CHECK ("status" = ANY (ARRAY['PENDING'::text, 'COMPLETED'::text, 'REJECTED'::text, 'EXPIRED'::text]))`
+        `ALTER TABLE "transactions" ADD CONSTRAINT "transactions_status_check" CHECK ("status" = ANY (ARRAY['PENDING'::text, 'COMPLETED'::text, 'REJECTED'::text, 'EXPIRED'::text, 'DELETED'::text]))`
```
dragosp1011, line 11: it should be a separate migration

dragosp1011: also a new migration should be added to mark as deleted the transactions linked to existent inactive wallet addresses

packages/wallet/backend/src/transaction/service.ts
```diff
@@ -87,6 +87,7 @@ export class TransactionService implements ITransactionService {
       )
       .fullOuterJoinRelated('[walletAddress, account.user]')
       .where('account:user.id', userId)
+      .whereNot('status', 'DELETED')
       .whereNotNull('transactions.id')
```
dragosp1011, line 91: as id is the primary key, curious why we have this check here
I think it should be deleted

packages/wallet/backend/src/walletAddress/service.ts
```diff
@@ -287,6 +289,11 @@ export class WalletAddressService implements IWalletAddressService {
     await WalletAddress.query().findById(id).patch({
       active: false
     })
+    console.log(walletAddress.accountId)
+    await this.transactionService.updateTransaction(
+      { accountId: walletAddress.accountId },
```
dragosp1011, line 294: it should be updated by wallet address id
there can be multiple wallet addresses in one account and it is not correct to mark as DELETED transactions that belongs to other wallet addresses that the one that is marked as inactive

dragosp1011 (changes requested):

packages/wallet/backend/src/walletAddress/service.ts
```diff
@@ -287,6 +289,11 @@ export class WalletAddressService implements IWalletAddressService {
     await WalletAddress.query().findById(id).patch({
       active: false
     })
+    console.log(walletAddress.accountId)
```
Tymmmy, line 292: remove console log

packages/wallet/backend/migrations/20230728101910_update_transactions_columns.js
dragosp1011: @rico191013 the transactions linked to inactive wallet addresses should be marked as DELETED in the migration

beniaminmunteanu (approved):
lgtm, but wait for dragos approval

packages/wallet/backend/migrations/20240215061440_add_delete_flag_to_transactions_status_check_constraint.js
```diff
@@ -0,0 +1,38 @@
...
+    await knex.raw(
+      [
+        `ALTER TABLE "transactions" DROP CONSTRAINT IF EXISTS "transactions_status_check";`,
+        `ALTER TABLE "transactions" ADD CONSTRAINT "transactions_status_check" CHECK ("status" = ANY (ARRAY['PENDING'::text, 'COMPLETED'::text, 'REJECTED'::text, 'EXPIRED'::text, 'DELETED'::text]))`
+      ].join('\n')
+    )
+
+    await knex.raw(
+      [
+        `UPDATE transactions`,
```
dragosp1011, line 16: a similar query should be added in `down` to restore the data.

dragosp1011: @beniaminmunteanu @Tymmmy @raducristianpopa 
I think it is not the best to add a new status in this case as the previous one is overwritten.
Keeping the status as it is and adding a separate field to indicate that transactions is deleted/inactive allows to have the correct transactions status for history and can be easily restored.

Tymmmy: Yes, I agree, we need a new field here to indicate that it is an inactive PP

beniaminmunteanu: good point.

packages/wallet/backend/migrations/20240215061440_add_active_column.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.alterTable('transactions', async (table) => {
+    table.boolean('active').notNullable().defaultTo(true)
```
dragosp1011, line 7: replace with `deletedAt` column that will keep the time of deletion

rico191013: How can I detect the time of deletion of old transactions?

dragosp1011: use `updatedAt` from wallet address

dragosp1011 (dismissed):

PR #1138, opened 2024-02-16
fix(wallet-frontend): Transaction view improvements

### LINKED ISSUE #543: Transactions view improvements
- [ ] Use a context for the filters container
- [x] Parse the query parameters with a zod schema
- [ ] Fetch the filters options (accounts and payment pointers) on the client side - remove GSSP
- [x] Button to clear all the filters
- [x] Better pagination component where we are going to display the page numbers as well (`Previous | 1 2 3 ... $lastPage | Next)`
- [x] Allow the user to specify the page size - now it has a default of 10
- [x] Allow ordering from newest to oldest and from oldest to newest by clicking on the `Date` column header
- [x] Use searchParams for transactions list api call

Tymmmy:
### Context

- fixes #543


packages/wallet/frontend/src/pages/transactions.tsx
```diff
@@ -269,32 +331,88 @@ const TransactionsPage: NextPageWithLayout<TransactionsPageProps> = ({
...
+                  ) {
+                    return (
+                      <li key={page} className="list-none p-1">
+                        <Button
+                          size="xs"
+                          intent="outline"
+                          className={cx(
+                            page - 1 === Number(pagination.page) &&
+                              '!bg-orange text-white'
+                          )}
```
raducristianpopa, line 375: Let's make the buttons green as well. I feel like the white-orange combination is a bit challenging for the eyes.

packages/wallet/frontend/src/lib/hooks/useTransactions.ts
```diff
@@ -20,9 +31,16 @@ const defaultState = {
 }
 
 export const useTransactions = () => {
-  const router = useRouter()
-  const { accountId, walletAddressId, type, status, page } =
-    router.query as TransactionsQueryParams
+  const router = useTypedRouter(transactionListQuerySchema)
```
raducristianpopa, line 34: We need to improve the `useTypedRouter` hook. Will have a look at it and get back.

raducristianpopa: https://github.com/interledger/testnet/blob/1f43d89fb50074b56a2552e9350f80d30968af55/packages/wallet/frontend/src/lib/hooks/useTypedRouter.ts#L5

```ts
import { type NextRouter, useRouter } from 'next/router'
import { z } from 'zod'

export function useTypedRouter<T extends z.ZodSchema>(
  schema: T
): NextRouter & { query: z.infer<T> } {
  const { query, ...router } = useRouter()

  return {
    query: schema.parse(query) as z.infer<typeof schema>,
    ...router
  }
}
```

This way we don't need to use type assertions when using the hook. The `list` function arguments need to be updated as well.

Zod's `parse` function will throw an error when validating, so we need to take this into consideration as well. Once solution is that we ignore invalid items and only return the valid ones, without throwing an error. For example, if `accountId` is not a `uuid`, we do not return it.

raducristianpopa (approved):

Issue #809, opened 2023-09-28
Add keyboard navigation to Testnet


Tymmmy:
We would like Testnet to be accessible by keyboard navigation only as well.
This applies to all Testnet components (wallet/boutique)

PR #1140, opened 2024-02-16
chore: added focus to the first focusable elements on component load

devcer:
### Context

- fixes https://github.com/interledger/testnet/issues/809

### Changes
* Added focus to the first focusable input element on component load across multiple components in wallet


Tymmmy:
We could add the same in these pages as well:
- forgotten password - focus on email input: 
image
- register/create account new user - focus on email input:
image


Maybe add these, then looks good :)

Tymmmy (changes requested):
Added some comments to add focus on 2 more elements.

Tymmmy (approved):

Issue #1136, opened 2024-02-15
rafiki.money vulnerability -> callbackURL parameter


Tymmmy:
We had some people from the community doing some testing in security vulnerabilities of testnet and we received the following suggestions:

Bug Title:
==========
Unvalidated Open Redirect and Reflected Cross-Site Scripting (XSS) in callbackUrl Parameter
Severity:
=========
High
Impact:
========
The unvalidated open redirect vulnerability combined with the reflected XSS in the callbackUrl parameter of the Rafiki Money authentication login page poses a significant security risk. Attackers can craft malicious URLs that redirect users to arbitrary websites and execute arbitrary JavaScript code in the context of the Rafiki Money domain. This can lead to various malicious activities, including phishing attacks, stealing sensitive user information, and unauthorized actions on behalf of users.
Reproduction Steps:
========================
Access the Rafiki Money authentication login page using the following URL: https://rafiki.money/auth/login?callbackUrl=https://google.com
Note the presence of the “callbackUrl” parameter in the URL.
Attempt to manipulate the “callbackUrl” parameter to include a malicious URL that redirects users to a different website while maintaining the appearance of a legitimate Rafiki Money login page.
Confirm that the redirect occurs without proper validation, allowing arbitrary redirection to external domains.
Similarly, attempt to inject XSS payloads into the “callbackUrl” parameter to execute arbitrary JavaScript code.
Confirm that the XSS payload gets reflected back in the response, potentially leading to script execution within the context of the Rafiki Money domain.
Proof of Concept (PoC):
========================
Access the following URL: https://rafiki.money/auth/login?callbackUrl=https://google.com
Note the redirection to Google’s homepage, indicating the presence of an unvalidated open redirect vulnerability.
Access the following URL to demonstrate the reflected XSS vulnerability:
[https://rafiki.money/auth/login?callbackUrl=JavaScript://%250Aalert?.(1)//'/*](https://rafiki.money/auth/login?callbackUrl=JavaScript://%250Aalert?.(1)//%27/*)\’/*“/*\“/*`/*\`/*%26apos;)/*
Note the execution of the alert message, indicating successful XSS exploitation.
Mitigation:
===========
1-   Implement proper input validation and sanitization mechanisms to prevent unvalidated open redirects and XSS vulnerabilities.
2-   Validate and restrict the “callbackUrl” parameter to only accept whitelisted URLs within the Rafiki Money domain.
[...]

PR #1213, opened 2024-03-12
fix: added valid url check in callback url redirection

devcer:
### Context

- fixes https://github.com/interledger/testnet/issues/1136

### Changes

Added an invalid URL check before the callback URL redirection

### DIFF
--- a/packages/wallet/frontend/src/pages/auth/login.tsx
+++ b/packages/wallet/frontend/src/pages/auth/login.tsx
@@ -32,7 +32,9 @@ const LoginPage: NextPageWithLayout = () => {
             const response = await userService.login(data)
 
             if (response.success) {
-              const isInvalidCallbackUrl = !callBackUrl.startsWith('/') && !callBackUrl.startsWith(window.location.origin);
+              const isInvalidCallbackUrl =
+                !callBackUrl.startsWith('/') &&
+                !callBackUrl.startsWith(window.location.origin)
               if (isInvalidCallbackUrl) {
                 router.push('/')
               } else {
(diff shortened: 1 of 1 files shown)


devcer:
@raducristianpopa @Tymmmy  The `new URL(...).searchParams.set(...)` already encode the URL, Hence encodeURIComponent is not needed in middleware. Otherwise, it is double encoding the already encoded URL, causing other URL issues.  To verify this, you could open `https://rafiki.money/transactions` in incognito and you will see the encoded callback param in the URL.

packages/wallet/frontend/src/pages/auth/login.tsx
```diff
@@ -32,7 +32,14 @@ const LoginPage: NextPageWithLayout = () => {
             const response = await userService.login(data)
 
             if (response.success) {
-              router.push(callBackUrl)
+              const isInvalidCallbackUrl =
+                !callBackUrl.startsWith('/') &&
+                !callBackUrl.startsWith(window.location.origin)
+              if (isInvalidCallbackUrl) {
```
Tymmmy, line 38: suggestion -> isIncorrectCallbackUrl ? router.push('/') : router.push(callbackUrl)

packages/wallet/frontend/src/pages/auth/login.tsx
```diff
@@ -32,7 +32,14 @@ const LoginPage: NextPageWithLayout = () => {
             const response = await userService.login(data)
 
             if (response.success) {
-              router.push(callBackUrl)
+              const isInvalidCallbackUrl =
```
Tymmmy, line 35:
```suggestion
              const isIncorrectCallbackUrl =
```

Tymmmy: Trying to use politically correct wording :D

Tymmmy (approved):

PR #1224, opened 2024-03-19
feat: Create new socket emit for occurred errors in rafiki webhook

### LINKED ISSUE #1209: Exchange money and Transactions problems for other currencies
Looks like on rafiki.money the Exchange money buttton and Transactions are getting EXPIRED when we are dealing with other currencies then USD or EUR.
Please check for example SGD, HKD, RON etc...

rico191013:
### Context

- fixes #1209

### DIFF
--- a/packages/wallet/backend/src/rafiki/service.ts
+++ b/packages/wallet/backend/src/rafiki/service.ts
@@ -12,7 +12,7 @@ import { WalletAddressService } from '@/walletAddress/service'
 import { WMTransactionService } from '@/webMonetization/transaction/service'
 import { Account } from '@/account/model'
 import { WMTransaction } from '@/webMonetization/transaction/model'
-import MessageType from '@/socket/messageType'
+import { MessageErrorType, MessageType } from '@/socket/messageType'
 
 export enum EventType {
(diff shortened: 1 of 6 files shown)


packages/wallet/backend/src/rafiki/service.ts
```diff
@@ -209,14 +209,33 @@ export class RafikiService implements IRafikiService {
...
-        } into ${receiverWalletId} error message: ${
-          transferResult.status?.message || 'unknown'
-        }`
-      )
+      switch (transferResult.status?.error_code) {
+        case MessageErrorType.ERROR_MINOR_CURRENCY_UNITS:
+          if (!user)
+            throw Error('User should not be undefined for emitting socket')
+
+          await this.socketService.emitError(
```
dragosp1011, line 220: webhook events will retry until the payment is completed.
User will be spammed with a lot of messages in this case.

Tymmmy: Yes, we will not be using notification boxes to display errors, we discussed with Rico, and only the logger will receive some more details about the error, the user will not be spammed with error messages.

PR #1229, opened 2024-03-23
fix: check for existing transaction

### LINKED ISSUE #1225: Boutique purchase retriggers transaction many time
Steps to reproduce:
- Go to your USD account in Testnet and make sure you have about 30$ (withdraw money if needed)
- Go to Boutique and buy Knowledge for example 54$ (make sure the price is greater then your USD funds)
- Approove the purchase, the purchase is successful
- Go to Testnet Transactions, and you will see 1 PENDING transaction for your Boutique purchase
- You don't have enough funds to Complete the previous purchase, but wait a few minutes before you top up
- After around 5 minutes add some more funds into your USD account
- Go to Transactions page
- You will see for every minute past since the purchase 2 transactions PENDING for Purchase at Boutique, then after the main transaction gets the money from the account, all of the become COMPLETED.

Before fund account we have many PENDING:
image
image

After funding they will all be COMPLETED:
image
image

How it should work :
- Only one transaction should be in the list for the one purchase
- Only one transaction should be PENDING before adding more funds
- After adding more funds, only that one transaction should be COMPLETED

dragosp1011:
### Context

- fixes #1225

### DIFF
--- a/packages/wallet/backend/src/transaction/service.ts
+++ b/packages/wallet/backend/src/transaction/service.ts
@@ -145,7 +145,9 @@ export class TransactionService implements ITransactionService {
     params: OutgoingPayment,
     walletAddress: WalletAddress
   ) {
-    const existentTransaction = await Transaction.query().findOne({ paymentId: params.id })
+    const existentTransaction = await Transaction.query().findOne({
+      paymentId: params.id
+    })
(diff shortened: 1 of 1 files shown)


Tymmmy (approved):
That's it :D easy fix :)

PR #1251, opened 2024-03-28
feat(shared): migrate configs into shared package

### LINKED ISSUE #1190: [Subtask] - Share config files
Share libraries between Testnet Wallet backend and Testnet Boutique backend.
Identified shareable locations:

config/logger
config/knex

rico191013:
### Context

- fixes #1190


packages/shared/backend/src/configs/env.ts
```diff
@@ -0,0 +1,17 @@
+import { z } from 'zod'
```
dragosp1011, line 1: env file cannot be moved in shared as each app has different variables

packages/shared/backend/src/configs/knex.ts
```diff
@@ -2,6 +2,7 @@ import knex from 'knex'
 import { Env } from './env'
 
 export function createKnex(env: Env) {
+  console.log('database_url', env.DATABASE_URL)
```
dragosp1011, line 5: remove console.log

packages/shared/backend/src/configs/knex.ts
```diff
@@ -2,6 +2,7 @@ import knex from 'knex'
 import { Env } from './env'
 
 export function createKnex(env: Env) {
```
dragosp1011, line 4: define an interface for required fields in knex config instead of importing Env

dragosp1011 (approved):

PR #1253, opened 2024-03-28
fix: start background processing only if the app started

### LINKED ISSUE #1252: Background jobs are started despite application not being ready
There are multiple errors on starting the app from scratch.
Boutique app does not start because the background job is trying to query the db that is not yet ready.
Same in wallet - there are multiple queries to db to process pending transactions. The wallet is eventually starting after the connection to db is established.

The background jobs should start only if the application started successfully.
image

dragosp1011:
### Context

- fixes #1252

### DIFF
--- a/packages/boutique/backend/src/app.ts
+++ b/packages/boutique/backend/src/app.ts
@@ -132,7 +132,7 @@ export class App {
       .catch((err) => {
         logger.error('Error while trying to process pending payments')
         logger.error(err)
-        return true
+        return false
       })
       .then((trx) => {
         if (trx) {
--- a/packages/boutique/backend/src/index.ts
+++ b/packages/boutique/backend/src/index.ts
@@ -1,8 +1,8 @@
 import { createContainer } from './container'
 import { env } from './config/env'
 import { App } from './app'
-import * as console from "console";
-import * as process from "process";
+import * as console from 'console'
+import * as process from 'process'
 
 export const start = async (app: App): Promise<void> => {
   await app.startServer()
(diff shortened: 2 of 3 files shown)


packages/boutique/backend/src/index.ts
```diff
@@ -9,10 +11,14 @@ export const start = async (app: App): Promise<void> => {
 async function bootstrap() {
   const container = await createContainer(env)
   const app = new App(container)
-  await start(app).catch(async (e): Promise<void> => {
+  try {
+    await start(app)
+    await app.processResources()
+  } catch (e) {
+    console.log('Error on starting the app')
```
Tymmmy, line 18: should we leave in console logs?

dragosp1011: yes, the app is not initialised on this step yet

packages/boutique/backend/src/app.ts
```diff
@@ -132,7 +132,7 @@ export class App {
       .catch((err) => {
         logger.error('Error while trying to process pending payments')
         logger.error(err)
-        return true
+        return false
```
beniaminmunteanu, line 135: why is the return needed at all?

dragosp1011: yes, to setup the next execution

dragosp1011: with returning true, it was going inside the if with immediate execution and flooding the app with errors

beniaminmunteanu (approved):

Tymmmy (approved):

PR #1269, opened 2024-04-02
feat(shared): migrate cache into shared package

### LINKED ISSUE #1191: [Subtask] - Share cache files
Share libraries between Testnet Wallet backend and Testnet Boutique backend.
Identified shareable locations:

cache - implement redis and in-memory also

rico191013:
### Context

- fixes #1191


packages/boutique/backend/src/cache/one-click.ts
```diff
@@ -8,4 +8,7 @@ export type OneClickCacheData = {
   continueToken: string
 }
 
-export type OneClickCache = InMemoryCache<OneClickCacheData>
+export type OneClickCache = Cache<OneClickCacheData>
```
dragosp1011, line 11: why the cache type was changed?

rico191013: I thought it would be better for all parts of the backends to use the same cache type at this case `OneClickCach` structure does not depend on `in-memory`  cache type so I decided to change it to the Redis type and it caused we don't worry about its expiration anymore so I cleared its expiration process in `app.ts`

dragosp1011: I think we should keep the same implementation. @raducristianpopa any thoughts?

packages/boutique/backend/src/app.ts
```diff
@@ -114,15 +114,7 @@ export class App {
...
-
-  private async processOneClickCache() {
-    const oneClickCache = this.container.resolve('oneClickCache')
-    oneClickCache.processExpired()
-    setTimeout(() => this.processOneClickCache(), 5000).unref()
-  }
-
   async processResources() {
     process.nextTick(() => this.processPendingPayments())
-    process.nextTick(() => this.processOneClickCache())
```
dragosp1011, line 126: why is this removed?

packages/boutique/backend/src/cache/one-click.ts
raducristianpopa: I agree with @dragosp1011. Using Redis for `OneClickCache` is not really necessary and we should stick with the in memory one.

dragosp1011 (approved):

PR #1284, opened 2024-04-10
feat(upgrade): upgrade to alpha 8

beniaminmunteanu:
This PR upgrades testnet to rafiki alpha-8
& upgrades node version to 20.12.1 (current iron lts)

### DIFF
--- a/docker/prod/docker-compose.yml
+++ b/docker/prod/docker-compose.yml
@@ -119,7 +119,7 @@ services:
     <<: *logging
 
   rafiki-auth:
-    image: ghcr.io/interledger/rafiki-auth:v1.0.0-alpha.7
+    image: ghcr.io/interledger/rafiki-auth:v1.0.0-alpha.8
     container_name: rafiki-auth
     environment:
       NODE_ENV: ${NODE_ENV}
@@ -143,7 +143,7 @@ services:
     <<: *logging
 
   rafiki-backend:
-    image: ghcr.io/interledger/rafiki-backend:v1.0.0-alpha.7
+    image: ghcr.io/interledger/rafiki-backend:v1.0.0-alpha.8
     container_name: rafiki-backend
     depends_on:
       - postgres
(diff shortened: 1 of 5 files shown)


Tymmmy (approved):

packages/boutique/backend/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "@google-cloud/logging-winston": "^6.0.0",
-    "@interledger/open-payments": "^6.7.1",
+    "@interledger/open-payments": "^6.8.0",
```
raducristianpopa, line 11: Is there an issue with the `6.8.2` version ?

docker/prod/docker-compose.yml
```diff
@@ -190,7 +190,7 @@ services:
     <<: *logging
 
   rafiki-frontend:
-    image: ghcr.io/interledger/rafiki-frontend:v1.0.0-alpha.7
+    image: ghcr.io/interledger/rafiki-frontend:v1.0.0-8
```
dragosp1011, line 193: alpha?

package.json
```diff
@@ -50,7 +50,7 @@
     "pnpm": "^8.15.6",
     "npm": "pnpm",
     "yarn": "pnpm",
-    "node": "^18.14.0"
+    "node": "^20.12.1"
```
dragosp1011, line 53: what about Docker files?

packages/boutique/backend/package.json
dragosp1011: probably not, this is the version in rafiki

beniaminmunteanu: will go for ^ 6.8.2 then, rafiki already uses 6.8.2 just that ^6.8.0 is in the package json

package.json
beniaminmunteanu: have another look please

dragosp1011 (approved):

dragosp1011 (approved):

PR #1287, opened 2024-04-10
feat(logger): add service name as context

dragosp1011:
### Context

- fixes #


raducristianpopa (commented):
Instead of creating the logger in every class, we can inject it directly with Awilix:

```ts
container.register({
    logger: asValue(logger),
    sessionService: asClass(SessionService)
      .singleton()
      .inject(() => ({
        logger: logger.child(SessionService.name)
      })),
 })
```

This solution works as well, but I think it fits better to have everything loaded by Awilix.

beniaminmunteanu (dismissed):

beniaminmunteanu (approved):

PR #1300, opened 2024-04-16
feat: implement shared zod schema between backend and frontend

### LINKED ISSUE #1188: Share libraries in Wallet FE <-> BE
Share libraries between testnet Wallet backend and frontends.
First look into complexity of implementing Open API.
Share Zod schemas.

rico191013:
### Context

- fixes #1188


package.json
```diff
@@ -9,7 +9,6 @@
     "url": "https://github.com/interledger/testnet"
   },
   "scripts": {
-    "shared:backend": "pnpm --filter @shared/backend --",
```
dragosp1011, line 12: why it was removed?

packages/boutique/backend/package.json
```diff
@@ -2,14 +2,15 @@
...
-    "build:deps": "pnpm --filter @shared/backend build",
+    "build:deps": "pnpm --filter @shared/backend build && pnpm --filter @shared/boutique build",
     "build": "pnpm build:deps && tsc --build tsconfig.build.json && tsc-alias -p tsconfig.build.json",
     "test": "jest --passWithNoTests --maxWorkers=2"
   },
   "dependencies": {
     "@google-cloud/logging-winston": "^6.0.0",
     "@interledger/open-payments": "^6.8.2",
     "@shared/backend": "workspace:*",
+    "@shared/boutique": "workspace:*",
```
dragosp1011, line 13: as it is only for boutique, it make sense to be part of boutique `boutique/shared`

packages/boutique/backend/package.json
```diff
@@ -2,14 +2,15 @@
   "name": "@boutique/backend",
   "scripts": {
     "start": "node dist/index.js",
-    "build:deps": "pnpm --filter @shared/backend build",
+    "build:deps": "pnpm --filter @shared/backend build && pnpm --filter @shared/boutique build",
```
dragosp1011, line 5: `@shared/boutique` -> `@boutique/shared`

packages/boutique/frontend/vite.config.ts
```diff
@@ -8,6 +8,9 @@ export default defineConfig({
     port: 4004
   },
   resolve: {
-    alias: [{ find: '@', replacement: resolve(__dirname, 'src') }]
+    alias: {
+      '@': resolve(__dirname, 'src'),
+      '@shared/boutique': resolve(__dirname, '../../boutique/shared/src')
```
raducristianpopa, line 13: No need to have the type alias if `@shared/boutique` is a package.

packages/boutique/shared/index.ts
raducristianpopa, line 1: Move `index.ts` in `src`.

packages/boutique/backend/Dockerfile.prod
```diff
@@ -24,10 +25,12 @@ WORKDIR /testnet
 COPY package.json pnpm-lock.yaml pnpm-workspace.yaml .npmrc .nvmrc tsconfig.base.json  ./
 COPY packages/boutique/backend ./packages/boutique/backend
 COPY packages/shared/backend ./packages/shared/backend
+COPY packages/boutique/shared ./packages/boutique/shared
 
 RUN pnpm boutique:backend build
 RUN rm -rf ./node_modules ./packages/boutique/backend/node_modules
 RUN rm -rf ./node_modules ./packages/shared/backend/node_modules
+RUN rm -rf ./node_modules ./packages/boutique/shared/node_modules
```
raducristianpopa, line 33: Why are we removing the root `node_modules` 3 times. Doing `RUN rm -rf ./node_modules ./packages/boutique/backend/node_modules ./packages/shared/backend/node_modules ./packages/boutique/shared/node_modules` is breaking something?

packages/boutique/frontend/vite.config.ts
rico191013: well without this we will get  `Failed to resolve entry for package "@boutique/shared"` error in locally run build by docker

beniaminmunteanu (commented):
If this is the structure we're aiming for:

- a shared package at the same level with wallet and boutique, that has backend shared functionality.

-  another shared directory both in wallet and boutique, which basically holds API types / schema.

I'd change the 'shared' directory name in wallet and boutique to 'models', so there is less naming confusions.

thoughts?

packages/boutique/frontend/vite.config.ts
raducristianpopa: I've looked into this. The main reason behind this error is that `vite` is expecting the source code to be ESM only. and we are bundling `@boutique/shared` as CJS. This should be okay for now, but I think we should discuss if we want to move everything to ESM. A lot of libraries are starting to be bundled for ESM only.
- https://github.com/wooorm/npm-esm-vs-cjs?tab=readme-ov-file#data

Node 22 will have support for importing ESM into CJS using `require`, but we will be able to use this functionality in Octomber/November.
- https://nodejs.org/en/blog/announcements/v22-release-announce#support-requireing-synchronous-esm-graphs
- https://nodejs.org/en/about/previous-releases

Converting the shared package to ESM will throw a runtime error when boutique's backend is ran:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/radu/dev/testnet/packages/boutique/shared/dist/index.js from /Users/radu/dev/testnet/packages/boutique/backend/dist/order/controller.js not supported.
Instead change the require of index.js in /Users/radu/dev/testnet/packages/boutique/backend/dist/order/controller.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/radu/dev/testnet/packages/boutique/backend/dist/order/controller.js:7:18) {
  code: 'ERR_REQUIRE_ESM'
}
```

packages/boutique/shared/index.ts
raducristianpopa: I've put up a PR that takes cares of this: https://github.com/interledger/testnet/pull/1343.

rico191013: Still, we have to change all imports
Screenshot 1403-02-18 at 7 19 30 PM

dragosp1011 (approved):

PR #1344, opened 2024-05-03
feat(1311): update transfer pages ux

### LINKED ISSUE #1311: Update Transfer section UX
## Current implementation

The following are screenshots of the send and request money screens.

Screenshot 2024-04-18 at 3 57 18 PM
_Send money screen_

Screenshot 2024-04-18 at 3 57 23 PM
_Request money screen_

They are the only 2 screens out of all the pages that are navigable from the sidebar that do not have a page title.

When an account is selected, the amount of funds in the account gets rendered under the "Total balance" but before any account is selected, there is nothing there, and no context provided to why there is a "Total balance" in the middle of nowhere. 

On the Request money screen, the label for the form uses the text "to", which can be confusing because in the Send money screen, we use "to" to indicate the receiver's account.

## Proposed change

There are a total of **3 proposed changes** to address the above 3 concerns:

1. If we could add a page title to the "Send" and "Request" pages, users could more easily see where they are. And if we use the same page title component as the other pages, we can remove the need to use `position: absolute` on the "Total balance" component.

    My understanding is that absolute positioning was used to ensure that the "Total balance" component would be perfectly centred on mobile, when the back button is visible. If we use the same page title component as all the other pages, the back button would then be part of the page title. We can then remove the back button from the "Total balance" component, and absolute positioning would not be needed at all.

2. If we could add an empty state of "-.--" before any account is selected under the text "Total balance", this could make things a little clearer (please see attached screenshots for a better idea of what this proposal entails)

3. If we could change the text on the label from "to" to "into", then it might help because then people could understand the process as requesting money _into_ their own account.

Screenshot 2024-04-18 at 4 20 27 PM
_Proposed Send page on desktop view_

Screenshot 2024-04-18 at 4 20 44 PM
_Proposed Send page on mobile view_

Screenshot 2024-04-23 at 1 57 42 PM
_Proposed Request page on desktop view_

Screenshot 2024-04-23 at 1 57 27 PM
_Proposed Request page on mobile view_

lengyel-arpad85:
In transfer pages, added pageHeader and value placeholder, adjusted display of transfer header, label update

### Context

- fixes #1311

### DIFF
--- a/packages/wallet/frontend/src/components/TransferHeader.tsx
+++ b/packages/wallet/frontend/src/components/TransferHeader.tsx
@@ -1,5 +1,4 @@
 import { cx } from 'class-variance-authority'
-import { BackButton } from './BackButton'
 
 const TYPES = {
   pink: 'text-pink',
@@ -17,18 +16,12 @@ type TransferHeaderProps = {
 export const TransferHeader = ({ type, balance }: TransferHeaderProps) => {
   return (
     <div className="text-center">
-      <div className="relative mb-16 mt-7 flex items-center">
-        <BackButton />
-        <div className="absolute left-[50%] -translate-x-1/2">
-          <h2 className={cx('text-xl font-light', TYPES[type])}>
-            Total balance
-          </h2>
-          {balance && (
-            <h3 className={cx('h-10 text-3xl font-semibold', TYPES[type])}>
-              {balance}
-            </h3>
-          )}
-        </div>
+      <div className="mb-16 mt-7 flex flex-col items-center justify-center">
+        <h2 className={cx('text-xl font-light', TYPES[type])}>Total balance</h2>
+
+        <h3 className={cx('h-10 text-3xl font-semibold', TYPES[type])}>
+          {balance ? balance : '-.--'}
+        </h3>
       </div>
     </div>
   )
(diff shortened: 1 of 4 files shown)


packages/wallet/frontend/src/components/TransferHeader.tsx
```diff
@@ -17,18 +16,12 @@ type TransferHeaderProps = {
...
-              {balance}
-            </h3>
-          )}
-        </div>
+      <div className="mb-16 mt-7 flex flex-col items-center justify-center">
+        <h2 className={cx('text-xl font-light', TYPES[type])}>Total balance</h2>
+
+        <h3 className={cx('h-10 text-3xl font-semibold', TYPES[type])}>
+          {balance ? balance : '-.--'}
+        </h3>
```
raducristianpopa, line 24: We need to keep the `BackButton` component for mobile.

lengyel-arpad85: @raducristianpopa PageHeader has it, if we keep it in TransferHeader also, it will be duplicated

raducristianpopa (approved):

Tymmmy (approved):

PR #1345, opened 2024-05-07
feat: implement open api

### LINKED ISSUE #1334: [Subtask] - Open API  for Account
Implement Open API for Testnet Wallet backend and Testnet Boutique backend
Controllers: 

account

rico191013:
Added Open api for accounts

### Context

- fixes #1334

### DIFF
--- a/packages/wallet/backend/src/open-api/config.ts
+++ b/packages/wallet/backend/src/open-api/config.ts
@@ -1,0 +1,41 @@
+import swaggerJsdoc from 'swagger-jsdoc'
+import { AccountPaths, AccountSchema } from '@/account/account.docs'
+const options = {
+  definition: {
+    openapi: '3.1.0',
+    info: {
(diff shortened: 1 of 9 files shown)


packages/wallet/backend/src/open-api/config.ts
```diff
@@ -0,0 +1,41 @@
...
+    info: {
+      title: 'Rafiki Wallet Express API with Swagger',
+      version: '0.1.0',
+      description:
+        'This is a simple CRUD API application made with Express and documented with Swagger',
+      license: {
+        name: 'MIT',
+        url: 'https://spdx.org/licenses/MIT.html'
+      },
+      contact: {
```
beniaminmunteanu, line 15: what is this?

rico191013: No We should fill it with valid contacts

PR #1355, opened 2024-05-21
feat: parse open-payments intervals

### LINKED ISSUE #1325: Grants detail page - make it better
We need to make a human readable string from the grant interval - Grants detail page when having a web monetization grant

https://github.com/tolu/ISO8601-duration?tab=readme-ov-file#example
https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals
We will probably need to parse this ourselves (maybe another library in the open-payments repo?)

image

rico191013:
### Context

- fixes #1325


packages/wallet/backend/src/grant/service.ts
```diff
@@ -80,6 +80,32 @@ export class GrantService implements IGrantService {
...
+          else if (months === 12)
+            accessElement.limits!.interval = `every year from ${date.format('ddd')}`
+          else if (months > 12) {
+            const years = duration.asYears()
+            accessElement.limits!.interval = `every ${years} year from ${date.format('ddd')}`
+          } else if (months < 1 && months > 0) {
+            const days = duration.asDays()
+            accessElement.limits!.interval = `every ${days} days from ${date.format('ddd')}`
+          }
+        }
```
beniaminmunteanu, line 109: How do you feel about an implementation of this method that does not mutate the grants list?
(by using .map)
(This would also imply the above changes for the usage of the function)
```suggestion
  private parseIntervals(grants: Grant[]): Grant[] {
  return grants.map(grant => {
    const access = grant.access.map(accessElement => {
      if (accessElement.limits?.interval) {
        const time = accessElement.limits.interval.split('/')
        const duration = moment.duration(time[2])
        const date = moment(time[1])
        const months = duration.asMonths()

        let interval: string;
        if (months === 1) {
          interval = `every month from ${date.format('ddd')}`;
        } else if (months > 1 && months < 12) {
          interval = `every ${months} month from ${date.format('ddd')}`;
        } else if (months === 12) {
          interval = `every year from ${date.format('ddd')}`;
        } else if (months > 12) {
          const years = duration.asYears();
          interval = `every ${years} year from ${date.format('ddd')}`;
        } else if (months < 1 && months > 0) {
          const days = duration.asDays();
          interval = `every ${days} days from ${date.format('ddd')}`;
        }

        return {
          ...accessElement,
          limits: {
            ...accessElement.limits,
            interval: interval,
          },
        };
      }

      return accessElement;
    });

    return { ...grant, access };
  });
}
```

packages/wallet/backend/src/grant/service.ts
```diff
@@ -80,6 +80,32 @@ export class GrantService implements IGrantService {
       args.filter = { identifier: { in: identifiers } }
     }
 
-    return await this.rafikiAuthService.listGrantsWithPagination(args)
+    const grants = await this.rafikiAuthService.listGrantsWithPagination(args)
+    this.parseIntervals(grants.grants.edges.map((it) => it.node))
+    return grants
```
beniaminmunteanu, line 85: Usage of the function if it is implemented without mutating grants directly:

```suggestion
   const grants = await this.rafikiAuthService.listGrantsWithPagination(args)
  const parsedGrants = this.parseIntervals(grants.grants.edges.map((it) => it.node))
  return { ...grants, grants: { ...grants.grants, edges: parsedGrants.map(grant => ({ node: grant })) } }
```

beniaminmunteanu, line 85: Usage of the non-mutating function would be sth like:
(Not tested)
```suggestion
   const grants = await this.rafikiAuthService.listGrantsWithPagination(args)
  const parsedGrants = this.parseIntervals(grants.grants.edges.map((it) => it.node))
  return {
    ...grants,
    grants: {
      ...grants.grants,
      edges: grants.grants.edges.map((edge, index) => ({ ...edge, node: parsedGrants[index] }))
    }
  }
```

packages/wallet/backend/src/grant/service.ts
beniaminmunteanu: If you find it decreases readability, I guess we could settle for mutating the grants in the parse method as well.
Up to you & the others

package.json
```diff
@@ -47,11 +47,11 @@
...
   "engines": {
-    "pnpm": "^9.1.1",
+    "pnpm": "^9.1.2",
     "npm": "pnpm",
     "yarn": "pnpm",
     "node": "^20.12.1"
   },
   "private": true,
-  "packageManager": "pnpm@9.1.1"
+  "packageManager": "pnpm@9.1.2"
```
beniaminmunteanu, line 56: doesn't deploy pipeline need to adapt to this as well?
Also i'd prefer keeping such upgrades outside of feature PR's like this one

rico191013: If I didn't make this change, CI would encounter an error that the pnpm version should be updated.

packages/wallet/backend/src/grant/service.ts
rico191013: It would be harder to read than the current one.
and it would have more time complexity because we have one more for loop here

packages/wallet/backend/src/grant/service.ts
```diff
@@ -80,6 +80,32 @@ export class GrantService implements IGrantService {
...
+          const months = duration.asMonths()
+          if (months === 1)
+            accessElement.limits!.interval = `every month from ${date.format('ddd')}`
+          else if (months > 1 && months < 12)
+            accessElement.limits!.interval = `every ${months} month from ${date.format('ddd')}`
+          else if (months === 12)
+            accessElement.limits!.interval = `every year from ${date.format('ddd')}`
+          else if (months > 12) {
+            const years = duration.asYears()
+            accessElement.limits!.interval = `every ${years} year from ${date.format('ddd')}`
```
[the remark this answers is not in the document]
rico191013: their messages are different!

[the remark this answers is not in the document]
beniaminmunteanu: yes, the choice is between having 2 branches or having string interpolation in just one.
I'm fine with both

packages/wallet/backend/src/grant/service.ts
beniaminmunteanu: anything that does not mutate data will implicitly have 1 extra loop

package.json
beniaminmunteanu: [ok i saw this](https://github.com/interledger/testnet/pull/1333/files#diff-dd667e3260f679fc2c4cb8dd11a19dc1775d6a9e24025a99ef96c2775861c6dfR31)

Should still be a separate PR for that merged b4 this one, but ok

beniaminmunteanu (approved):

sanducb (dismissed):

packages/wallet/backend/src/grant/service.ts
```diff
@@ -80,6 +80,32 @@ export class GrantService implements IGrantService {
...
+    const grants = await this.rafikiAuthService.listGrantsWithPagination(args)
+    this.parseIntervals(grants.grants.edges.map((it) => it.node))
+    return grants
+  }
+
+  private parseIntervals(grants: Grant[]) {
+    for (const grant of grants)
+      for (const accessElement of grant.access)
+        if (accessElement.limits?.interval) {
+          const time = accessElement.limits?.interval.split('/')
```
dragosp1011, line 92: extract logic to transform interval in human readable format in a separate method

dragosp1011, line 92: why the number of repetition is ignored? isn't this piece important to show to the user

packages/wallet/backend/src/grant/service.ts
```diff
@@ -80,6 +80,32 @@ export class GrantService implements IGrantService {
...
+          const duration = moment.duration(time[2])
+          const date = moment(time[1])
+          const months = duration.asMonths()
+          if (months === 1)
+            accessElement.limits!.interval = `every month from ${date.format('ddd')}`
+          else if (months > 1 && months < 12)
+            accessElement.limits!.interval = `every ${months} month from ${date.format('ddd')}`
+          else if (months === 12)
+            accessElement.limits!.interval = `every year from ${date.format('ddd')}`
+          else if (months > 12) {
```
dragosp1011, line 102: the implementation looks to be focused on well defined intervals.
for the interval `R5/2008-03-01T13:00:00Z/P1Y2M10DT2H30M`
it will be displayed that it repeats every year, right? that will give the user incorrect info

packages/wallet/backend/tests/mocks.ts
```diff
@@ -308,12 +308,50 @@ export const mockedListGrant = [
...
+            value: '12425',
+            assetCode: 'USD',
+            assetScale: 2
+          },
+          receiveAmount: {
+            value: '12300',
+            assetCode: 'USD',
+            assetScale: 2
+          },
+          interval: 'R/2016-08-23T08:00:00Z/P1M'
```
dragosp1011, line 327: why the mocks are changed but there are no tests to check the interval transformation?

rico191013: It covers the list pagination test

packages/wallet/backend/src/grant/service.ts
dragosp1011: @rico191013 this comment was not handled

packages/wallet/backend/src/grant/service.ts
```diff
@@ -80,6 +80,41 @@ export class GrantService implements IGrantService {
...
+            accessElement.limits!.interval = `${isEnd ? 'Until' : 'From'} ${date.format('MMMM Do YYYY')} with no repetition`
+          else if (repetition === undefined)
+            accessElement.limits!.interval = `${this.processDuration(duration)} ${isEnd ? 'until' : 'from'} ${date.format('MMMM Do YYYY')}`
+          else
+            accessElement.limits!.interval = `${repetition} times ${this.processDuration(duration)} ${isEnd ? 'until' : 'from'} ${date.format('MMMM Do YYYY')}`
+        }
+  }
+
+  private processDuration(duration: moment.Duration) {
+    const years = duration.years() !== 0 ? `${duration.years()} years, ` : ''
```
dragosp1011, line 110: add a separate check if the value is 1 for the text to be on singular

packages/wallet/backend/src/grant/service.ts
```diff
@@ -80,6 +80,41 @@ export class GrantService implements IGrantService {
       args.filter = { identifier: { in: identifiers } }
     }
 
-    return await this.rafikiAuthService.listGrantsWithPagination(args)
+    const grants = await this.rafikiAuthService.listGrantsWithPagination(args)
+    this.parseIntervals(grants.grants.edges.map((it) => it.node))
```
dragosp1011, line 84: this type of manipulation is confusing.
I propose to change it to something like:
```
grants.grants.edges.forEach((edge) =>  {
  edge.access = this.parseAccessIntervals(edge.access)
})

.....

parseAccessIntervals(accesses) {
  return accesses.map((access) => {
    if ((access.limits?.interval) {
      return {
        ...access,
        limits: {
          ...access.limits,
          intervalHR: transformIntervaltoHR(access.limits.interval)
        }
      }
    }
    return access
  })
}
```

dragosp1011 (approved):

Issue #1351, opened 2024-05-13
add response types along with the shared zod schemas


beniaminmunteanu:
The testnet frontends would benefit from having the response types of the backend APIs in a shared directory, so that they can be referenced instead of duplicated.

So the same way that we are sharing the zod schemas for request validations, we should also share the response types.

Suggestion:
Maybe write them as interfaces, and have the backend Models also implement that interface

Controllers: 
- [x]  [wallet/account](https://github.com/interledger/testnet/issues/1334)
- [x] [wallet/asset](https://github.com/interledger/testnet/issues/1335)
- [x] [wallet/user](https://github.com/interledger/testnet/issues/1433)
- [x] [wallet/grant](https://github.com/interledger/testnet/issues/1337)
- [x] [wallet/incomingPayment](https://github.com/interledger/testnet/issues/1340)

PR #1361, opened 2024-06-06
feat: bring asset response type into shared package

### LINKED ISSUE #1335: [Subtask] - Share types for Asset
Share types for Testnet Wallet backend and Testnet Boutique backend
Controllers:

asset

rico191013:
### Context

- fixes #1335


packages/wallet/shared/src/types/asset.ts
```diff
@@ -0,0 +1,19 @@
+export type Asset = {
```
dragosp1011, line 1: are these types used in backend as well?

beniaminmunteanu: #1351 
 The Types shared here should be imported in the backend to be used as the response type of controllers, 
and also in the frontend in order to import and not duplicate response types.

So i believe these types should also be added to the controllers?

beniaminmunteanu (dismissed):

packages/wallet/shared/src/types/asset.ts
```diff
@@ -0,0 +1,12 @@
+export interface Asset {
+  id: string
+  code: string
+  scale: number
+  withdrawalThreshold?: bigint
+  createdAt: string
+}
+
+export interface Rates {
```
dragosp1011, line 9: move rates type in a separate file

beniaminmunteanu:
Keeping in mind the other places that need to have the types moved to shared: #1351 

My suggestion would be to have the interfaces in shared have a conventional name, like including "Response" in it such as "NameResponse"

So that in the "account" implementation for example we would have the "Account" type being used for the bussiness logic inside the backend, and that is not shared, but for the controller response type we would have "AccountResponse"

Wdyt?

dragosp1011 (dismissed):

dragosp1011 (approved):

PR #1398, opened 2024-06-18
feat: add theme switcher

raducristianpopa:
### Context

- fixes #


.prettierrc.js
```diff
@@ -1,12 +1,11 @@
...
   printWidth: 80,
   tabWidth: 2,
   semi: false,
   singleQuote: true,
   quoteProps: 'consistent',
   jsxSingleQuote: false,
-  trailingComma: 'none',
-  plugins: ['prettier-plugin-tailwindcss'],
-  tailwindConfig: './packages/wallet/frontend/tailwind.config.js'
+  trailingComma: 'none'
```
raducristianpopa, line 10: Disabling it for now. Running Prettier will format all the files where we have the previous Tailwind classes (like `text-green-3`) since they are not present in the config.

Tymmmy (approved):

PR #1427, opened 2024-06-27
WM on Test wallet

### LINKED ISSUE #1357: [FE] Asset scale 2 to asset scale 9
- [ ] Getting rid of WM tab - every wallet address will be set to scale 9, display all in one tab
- [ ] Display as asset scale 2 below or hover for asset scale 9
- [ ] The transfer page should allow viewing asset scale 9 -> show asset scale 9s in list of send

Related to #1358, #1359.

### LINKED ISSUE #1358: [BE] Asset scale 2 to asset scale 9
- [ ] Convert transactions with different asset scales by using Rafiki rate service (leveraging quote)
- [ ] All transactions should take virtual balance into account

Changes required by #1357

sanducb:
### Context

- closes #1358
- closes #1357

### Changes

**Note**: Still WIP, more changes to be added

- Allowing creation of asset scale 9 wallet addresses by default
- Deleted WM accounting logic, wallet addresses are WM ready by default
- Added virtual balance logic for dealing with scale 9 amounts


docker/dev/docker-compose.yml
```diff
@@ -54,7 +54,7 @@ services:
       RATE_API_KEY: ${RATE_API_KEY}
       BASE_ASSET_SCALE: 2
       MAX_ASSET_SCALE: 9
-      WM_THRESHOLD: 100000000
+      RAPYD_THRESHOLD: 10000000 # 0.01
```
sanducb, line 57: I modified this to `0.01` (which is the lowest amount non-zero representable in asset scale 2) instead of `0.1`. Any reason to change it back?

packages/wallet/frontend/src/components/cards/AccountCard.tsx
```diff
@@ -19,10 +19,20 @@ export const AccountCard = ({ account, idOnboarding }: AccountCardProps) => {
...
+    const baseAssetScale = 2
+    const maxAssetScale = 9
+
+    const snapshotBalance = snapshotAccount
+      ? Number(snapshotAccount.balance)
+      : 0
+    const accountBalance =
+      Number(account.balance) * Math.pow(10, maxAssetScale - account.assetScale)
+    const value = (snapshotBalance || accountBalance).toString()
+
```
sanducb, line 31: This could be done differently/cleaner but haven't put too much effort into it since the whole interface will be changed.

packages/wallet/frontend/src/pages/account/[accountId].tsx
```diff
@@ -43,15 +39,38 @@ const AccountPage: NextPageWithLayout<AccountPageProps> = ({
...
+    const maxAssetScale = 9
+
+    const snapshotBalance = snapshotAccount
+      ? Number(snapshotAccount.balance)
+      : 0
+    const accountBalance =
+      Number(account.balance) * Math.pow(10, maxAssetScale - account.assetScale)
+
+    // `balance` represents incoming amount - outgoing amount in asset scale 9
+    const value = ((snapshotBalance || accountBalance) + balance).toString()
```
Comment (sanducb) [inline packages/wallet/frontend/src/pages/account/[accountId].tsx:57]: Same as here.

Tymmmy:
Something has happened with the quotes:
image

beniaminmunteanu:
Regarding question nr 1.:
Caching Wallet Addresses that are in use for WM is still a good ideea for latency, since in WM cases, we end up fetching the wallet address with a much higher frequency.
Since we would not have anything that differentiates Wallet addresses that are in use in WM to ones that are not, 
I think an LRU cache would achieve the same goal here?

On nr 2 and 3, tbh I don't know how those fields are used in order to answer.

Thoughts?
@dragosp1011  @raducristianpopa

Tymmmy:
Fee still is 0.... :(
image

sanducb:
I'll take another look at the frontend to make sure that every amount is properly formatted with a single shared function instead of the already existing code that I added for doing that which is not clean. I'll update the PR with the refactor.

dragosp1011 (changes requested):

packages/wallet/backend/migrations/20240801155211_update_wallet_addresses.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+  await knex('walletAddresses').update({ assetScale: 9 })
```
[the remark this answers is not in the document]
beniaminmunteanu: basically the assets in rafiki that had these wallet addresses linked to them should also be migrated to scale 9, right?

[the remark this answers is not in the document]
dragosp1011: yes, but as you cannot have multiple assets with the same code and scale, most likely the WA should be connected to the existent asset with scale 9

packages/wallet/backend/src/walletAddress/service.ts
```diff
@@ -159,62 +126,34 @@ export class WalletAddressService implements IWalletAddressService {
...
+      .where({ active: true })
       .joinRelated('account')
       .where({
         'account.userId': userId
       })
   }
 
   async getById(args: GetWalletAddressArgs): Promise<WalletAddress> {
-    //* Cache only contains WalletAddresses with isWM = true
-    const cacheHit = await this.cache.get(args.walletAddressId)
```
dragosp1011, line 212: as any WA address can be used for WM, should all WA be cached instead of removing it?

sanducb: I thought about this too, but I was curious if there's a reason why all addresses weren't already cached in the first place? If there isn't any, then I'm fine with caching all WA.

rico191013:
What will happen if we cache a WA, then update it, and then fetch it again before its expiry time has passed?

dragosp1011:
> What will happen if we cache a WA, then update it, and then fetch it again before its expiry time has passed?

there are 2 options is this case: 1. invalidate the cache on update or 2. cache the WA after update

packages/wallet/backend/src/walletAddress/service.ts
dragosp1011: the main reason why only WM WA where cached was because there are a lot of web monetization transactions in a short period of time. On simple WA there was no real reason to cache as the transactions will be dispersed in time.

As after these changes any WA can be used for WM it make sense to have the same caching mechanism for all WAs.

rico191013:
@sanducb please be aware that change the scale [here](https://github.com/interledger/testnet/pull/1484/files#) too.

packages/wallet/backend/src/rapyd/rapyd-client.ts
```diff
@@ -289,7 +289,7 @@ export class RapydClient implements IRapydClient {
       : ','
     const [street, city, postCode] = args.user.address?.split(', ') ?? []
     let address = [street, postCode].join(addressDelimiter)
-    if (args.assetCode === 'EUR') address = address.replace(/[,\s]+/g, '')
+    if (args.assetCode === 'USD') address = address.replace(/[,\s]+/g, '')
```
beniaminmunteanu, line 292: not sure why this transformation happens here tbh, but why was the asset code change needed?

packages/wallet/backend/src/walletAddress/service.ts
```diff
@@ -387,6 +345,7 @@ export class WalletAddressService implements IWalletAddressService {
 
     if (transfer.status?.status !== 'SUCCESS') {
       if (type === 'OUTGOING') {
+        //TODO this might not be needed
```
beniaminmunteanu, line 348: as in debt being part of the walletAddress model?

packages/wallet/frontend/src/pages/index.tsx
```diff
@@ -124,28 +131,47 @@ const HomePage: NextPageWithLayout<HomeProps> = ({ accounts, user }) => {
...
-  const response = await accountService.list(ctx.req.headers.cookie)
-  const user = await userService.me(ctx.req.headers.cookie)
+  const accountsResponse = await accountService.list(ctx.req.headers.cookie)
+  const userResponse = await userService.me(ctx.req.headers.cookie)
 
-  if (!response.success || !user.success) {
+  if (
+    !accountsResponse.success ||
+    !userResponse.success ||
+    !accountsResponse.result
```
beniaminmunteanu, line 147: At a quick glance, `accountResponse.result` might end up being an empty list:
image

This might not be the desired outcome of the condition^^
Double check if this can be the case, and if such, check for `accountsResponse.result.length` to reach desired condition

packages/wallet/frontend/src/pages/index.tsx
```diff
@@ -124,28 +131,47 @@ const HomePage: NextPageWithLayout<HomeProps> = ({ accounts, user }) => {
...
     }
   }
 
+  const incomingOutgoingBalances: Record<string, number> = {}
+  for (const account of accountsResponse.result) {
+    const walletAddressesResponse = await walletAddressService.list(
+      account.id,
+      ctx.req.headers.cookie
+    )
+    if (walletAddressesResponse.success && walletAddressesResponse.result) {
```
beniaminmunteanu, line 160: ditto above

README.md
```diff
@@ -91,7 +91,7 @@ The `RAPYD_ACCESS_KEY` and `RAPYD_SECRET_KEY` variables values can be found in y
 To create a new Interledger Test Wallet account, a verification email will be sent to the provided email address. If you want to send emails within the development environment, you will need to have a personal Sendgrid account and update the following environment variables: `SEND_EMAIL` to `true`, `SENDGRID_API_KEY` and `FROM_EMAIL`. If you prefer not to send emails in the development environment, simply set `SEND_EMAIL` to `false` and use the verification link found in the Docker `wallet-backend` container logs to finalize the registration process for a new user.
 
 Cross-currency transactions are supported. To enable this functionality, you will need to register at [freecurrencyapi.com/](https://freecurrencyapi.com/) and update the `RATE_API_KEY` environment variable with your own API key.
-Currencies can be added in the `admin` environment. For example `assetCode` is `EUR`, `assetScale` is `2`, and you will need to add an amount to `liquidity`.
+Currencies can be added in the `admin` environment. For example `assetCode` is `USD`, `assetScale` is `2`, and you will need to add an amount to `liquidity`.
```
Tymmmy, line 94: Why was it important to change EUR to USD?

packages/wallet/backend/src/rapyd/rapyd-client.ts
Tymmmy: Also interested in the answer here, we are creating default USD account for every new user in Rapyd, but don't know if that ihas anything to do with this. But why change EUR?

packages/wallet/backend/tests/mocks.ts
```diff
@@ -526,6 +526,6 @@ export const mockWalletAddress = {
 }
 
 export const mockExternalPayment = {
-  receivedAmount: { value: '0', assetCode: 'EUR', assetScale: 1 },
+  receivedAmount: { value: '0', assetCode: 'USD', assetScale: 1 },
```
Tymmmy, line 529: Why the change beetween assets?

packages/wallet/backend/tests/incomingPayment/service.test.ts
```diff
@@ -168,7 +168,7 @@ describe('Incoming Payment Service', () => {
         faker.internet.url()
       )
       expect(receivedAmount).toMatchObject({
-        assetCode: 'EUR',
+        assetCode: 'USD',
```
Tymmmy, line 171: Same question :D why the change?

Tymmmy (approved):

packages/wallet/backend/src/rapyd/rapyd-client.ts
sanducb: When I first started working on this issue I couldn't create an EUR account in the onboarding and thought that it should be USD by default. This was a change that I introduced in one of the first commits. Since then, I realised that I was wrong about the onboarding but forgot to revert the changes. Now I switched back to EUR 👍

Tymmmy (approved):

packages/wallet/frontend/src/pages/transfer/request.tsx
```diff
@@ -314,20 +329,28 @@ type SelectAccountOption = SelectOption &
...
+  if (!accountsResponse.success || !accountsResponse.result) {
     return {
       notFound: true
     }
   }
 
-  if (!accountsResponse.result) {
-    return {
-      notFound: true
+  const incomingOutgoingBalances: Record<string, number> = {}
```
sanducb, line 344: Should we share incoming/outgoing balance state so that we don't calculate it in each page as we do here (for example)?

packages/wallet/backend/knexfile.js
```diff
@@ -20,6 +20,20 @@ module.exports = {
     }
   },
 
+  rafiki_backend: {
```
dragosp1011, line 23: why is this added?

sanducb: I added this so that I can migrate assets to scale 9 in the `assets` table which is in `rafiki_backend` db.

packages/wallet/backend/migrations/20240801155211_update_asset_scale.js
```diff
@@ -0,0 +1,112 @@
...
+    'accountId'
+  )
+  for (const walletAddress of walletAddresses) {
+    const account = await knex('accounts')
+      .where('id', walletAddress.accountId)
+      .first()
+    if (account) {
+      await knex('walletAddresses')
+        .where('id', walletAddress.id)
+        .update({ assetCode: account.assetCode })
```
dragosp1011, line 51: why would asset code change?

packages/wallet/backend/knexfile.js
dragosp1011: add an env variable instead with the connection string

packages/wallet/backend/migrations/20240801155211_update_asset_scale.js
sanducb: This happens because `assetCode` was `null` for every WA for some reason and that was causing `keepBalancesSynced` to fail when checking that asset code exists for every WA. Until now, we only populated `assetCode` column on WAs table for WM payment pointers.

packages/wallet/backend/src/walletAddress/service.ts
sanducb: Yes, but I'm not 100% sure we should delete it

beniaminmunteanu: The point you are raising is valid. I think there should be a retry mechanism in place.

But I think the retry mechanism can be part of the current flow:
Transfer to rapyd fails -> transfer should be picked up by the next iteration of the cron-job and retried, since it should again be detected as imbalance.

This way, no additional mechanism requiring debt would be needed.

Tymmmy:
We will cherry pick the useful code we need from this.
Integrating Gatehub sandbox and replacing Rapyd, sadly made this story n0t needed anymore, as only asset code 2 will be supported.

PR #1446, opened 2024-07-28
feat: revoke recurring grants

### LINKED ISSUE #1444: Remove orders that related to revoked grants
Should remove orders from the one-click payment revoked their grants by users.

rico191013:
### Context

- fixes #1444


Tymmmy (dismissed):

packages/boutique/backend/src/order/controller.ts
```diff
@@ -222,10 +222,11 @@ export class OrderController implements IOrderController {
     res: TypedResponse<CreateResponse>,
     next: NextFunction
   ) => {
+    let order: Order
     try {
```
beniaminmunteanu, line 226: I'm not sure I get this right, but is this just a delete of the orderService in case the creation did not work due to having the OP client unauthorized?

If that is the case, Couldn't this just be a transaction that is rolled back, instead of a new specific delete call?

rico191013: Well, I tried it out. First, I moved the `instantBuy` method into the transaction block, but it didn’t work. It seemed that the order created in the transaction block wasn’t fully completed, which caused some database errors in the `instantBuy` method. So, I had to implement it this way instead.

beniaminmunteanu: Ok, if you're certain about it.
However, maybe in this case the Order.transaction(){} blocks are no longer needed, just 
try{
orderService.create()
}
catch{
orderService.delete()
}
?

beniaminmunteanu (approved):

PR #1461, opened 2024-07-31
fix: fix upload repetitive key into the Rafiki

### LINKED ISSUE #1460: Developer Keys duplicate keys uploade bug
## Bug Report

**Describe the bug**

Uploading the same key multiple times gives you correct error message, that you need to upload a unique one, but the backend API for uploading the key is still being called.

**To Reproduce**

- Upload a key in Developer Keys on Test Wallet, and check the API call.
- Upload the same key with a different nickname, and you will see the error message to upload a unique key, but check the API call again. The key is being uploaded.

**Expected behavior**

When a duplicate key is being uploaded backend API call for upload should not be called.

rico191013:
### Context

- fixes #1460

### DIFF
--- a/packages/wallet/backend/src/walletAddressKeys/service.ts
+++ b/packages/wallet/backend/src/walletAddressKeys/service.ts
@@ -6,7 +6,6 @@ import { WalletAddressKeys } from './model'
 import { WalletAddressService } from '@/walletAddress/service'
 import { WalletAddress } from '@/walletAddress/model'
 import { BadRequest, NotFound } from '@shared/backend'
-import { UniqueViolationError } from 'objection'
 
 export type KeyResponse = {
   privateKey: string
@@ -78,6 +77,15 @@ export class WalletAddressKeyService implements IWalletAddressKeyService {
       .export({ type: 'spki', format: 'pem' })
       .toString()
 
+    const isExist = await WalletAddress.query().where({
+      id: jwk.kid.toString()
+    })
+
+    if (isExist)
+      throw new BadRequest(
+        'Same key already uploaded. Please upload a unique one.'
+      )
+
     const walletAddressKey =
       await this.rafikiClient.createRafikiWalletAddressKey(
         jwk,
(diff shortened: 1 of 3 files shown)


packages/wallet/backend/src/walletAddressKeys/service.ts
```diff
@@ -78,6 +77,15 @@ export class WalletAddressKeyService implements IWalletAddressKeyService {
       .export({ type: 'spki', format: 'pem' })
       .toString()
 
+    const isExist = await WalletAddress.query().where({
```
Tymmmy, line 80:
```suggestion
    const isUploaded = await WalletAddress.query().where({
```

packages/wallet/backend/src/walletAddressKeys/service.ts
```diff
@@ -78,6 +77,15 @@ export class WalletAddressKeyService implements IWalletAddressKeyService {
...
+      id: jwk.kid.toString()
+    })
+
+    if (isExist)
```
Tymmmy, line 84:
```suggestion
    if (isUploaded)
```

packages/wallet/frontend/src/pages/grant-interactions/index.tsx
```diff
@@ -175,7 +175,7 @@ export const getServerSideProps: GetServerSideProps<{
     if (access.limits) {
       access.limits.receiver = access.limits.receiver
         ? replaceWalletAddressProtocol(access.limits.receiver)
-        : (access.limits.receiver ?? null)
+        : access.limits.receiver ?? null
```
Tymmmy, line 178: I think new prettier needs the ()

Tymmmy (approved):

dragosp1011 (approved):

dragosp1011 (changes requested):
please fix failed tests

dragosp1011 (approved):

Tymmmy (approved):

PR #1526, opened 2024-09-03
fix: remove WM accounting

### LINKED ISSUE #1507: [GateHub]-Remove Web monetization PPs from Test Wallet - new design
### Context

Cherry pick from https://github.com/interledger/testnet/pull/1427 the removal of WM PPs, and useful code from PR.

### Todos

- [ ] remove FE and BE part of WMs PPs
- [ ] add to env variables asset scale number

sanducb:
### Context

- closes #1507.

Since WM now works on asset scale 2 by default, WM accounting needs to be removed from test wallet because all WAs are WM enabled by default.

### Changes

- Removal of WM accounting
- Migration to update WAs with their respective account's asset code
- Addition of environment variables for asset scales instead of hardcoded values
- Formatting amounts differently so that they do not get ceiled
- Cache on update for WAs.


packages/wallet/backend/src/account/service.ts
```diff
@@ -244,7 +246,10 @@ export class AccountService implements IAccountService {
       accountId: existingAccount.id,
       paymentId: transactions[transactions.length - 1].id,
       assetCode: existingAccount.assetCode,
-      value: transformBalance(args.amount, asset.scale),
+      value: transformBalance(
```
dragosp1011, line 249: I believe here `asset.scale` should be kept for transforming.

packages/wallet/backend/src/account/service.ts
```diff
@@ -304,7 +309,9 @@ export class AccountService implements IAccountService {
     name = 'USD Account'
   ): Promise<Account | undefined> {
     const asset = (await this.rafikiClient.listAssets({ first: 100 })).find(
-      (asset) => asset.code === 'USD' && asset.scale === 2
+      (asset) =>
+        (asset.code === 'USD' && asset.scale === this.env.BASE_ASSET_SCALE) ||
+        undefined
```
dragosp1011, line 314: why `undefined` is used here?

packages/wallet/backend/src/walletAddress/service.ts
```diff
@@ -159,56 +124,38 @@ export class WalletAddressService implements IWalletAddressService {
         publicName: args.publicName,
         accountId: args.accountId,
         id: rafikiWalletAddress.id,
-        isWM: args.isWM,
-        assetCode: webMonetizationAsset?.code,
-        assetScale: webMonetizationAsset?.scale
+        assetCode: account.assetCode,
+        assetScale: this.env.BASE_ASSET_SCALE
```
dragosp1011, line 128: set scale from account

packages/wallet/backend/src/walletAddress/service.ts
```diff
@@ -159,56 +124,38 @@ export class WalletAddressService implements IWalletAddressService {
...
-      .where({ isWM: false, active: true })
+      .where({ active: true })
       .joinRelated('account')
       .where({
         'account.userId': userId
       })
   }
 
   async getById(args: GetWalletAddressArgs): Promise<WalletAddress> {
-    //* Cache only contains WalletAddresses with isWM = true
```
dragosp1011, line 211: as the method is performing additional validation on user and account, maybe caching should not be used checked in this case

packages/wallet/backend/migrations/20240903125600_update_wallet_addresses.js
```diff
@@ -0,0 +1,44 @@
...
+  for (const walletAddress of walletAddresses) {
+    const account = await knex('accounts')
+      .where('id', walletAddress.accountId)
+      .first()
+    if (account) {
+      await knex('walletAddresses')
+        .where('id', walletAddress.id)
+        .update({ assetCode: account.assetCode })
+    }
+  }
```
raducristianpopa, line 22: Why is this migration needed? Wouldn't an already existing wallet address have the same asset code as its account?

packages/wallet/backend/src/account/service.ts
raducristianpopa: I agree, no need to have a manual fallback to undefined. `find` will fallback to `undefined` by itself if the item is not found.

packages/wallet/frontend/src/components/dialogs/CreateWalletAddressDialog.tsx
```diff
@@ -67,13 +63,12 @@ export const CreateWalletAddressDialog = ({
                   as="h3"
                   className="text-center text-2xl font-bold"
                 >
-                  Create {titleExtra} Payment Pointer
+                  Create Payment Pointer
```
raducristianpopa, line 66: The `TemporaryWMNotice` component can be removed as well:
https://github.com/interledger/testnet/blob/1aaf8ea3092de5dc1e3fc913a8e34ecf75968a56/packages/wallet/frontend/src/components/dialogs/CreateWalletAddressDialog.tsx#L65-L68

packages/wallet/frontend/src/pages/account/[accountId].tsx
```diff
@@ -21,24 +21,15 @@ import { useEffect } from 'react'
...
 
 type AccountPageProps = InferGetServerSidePropsType<typeof getServerSideProps>
 
-const tabs = ['Account', 'Web Monetization']
-
 const AccountPage: NextPageWithLayout<AccountPageProps> = ({
   account,
-  allWalletAddresses,
-  balance
+  allWalletAddresses
```
Comment (raducristianpopa) [inline packages/wallet/frontend/src/pages/account/[accountId].tsx:32]:
```suggestion
  walletAddresses
```

packages/wallet/frontend/src/pages/account/[accountId].tsx
```diff
@@ -58,162 +49,99 @@ const AccountPage: NextPageWithLayout<AccountPageProps> = ({
...
+            </span>
+          </Link>
+        </div>
+        <div className="flex items-center justify-between">
+          <h3 className="text-lg font-semibold leading-none text-green">
+            Payment Pointers
+          </h3>
+        </div>
+        <div className="flex flex-col">
+          {allWalletAddresses.length > 0 ? (
```
Comment (raducristianpopa) [inline packages/wallet/frontend/src/pages/account/[accountId].tsx:133]:
```suggestion
          {walletAddresses.length > 0 ? (
```

packages/wallet/frontend/src/pages/account/[accountId].tsx
```diff
@@ -58,162 +49,99 @@ const AccountPage: NextPageWithLayout<AccountPageProps> = ({
...
+            <WalletAddressesTable
+              account={account}
+              walletAddresses={allWalletAddresses}
```
Comment (raducristianpopa) [inline packages/wallet/frontend/src/pages/account/[accountId].tsx:136]:
```suggestion
              walletAddresses={walletAddresses}
```

packages/wallet/frontend/src/pages/account/[accountId].tsx
```diff
@@ -224,8 +152,7 @@ const querySchema = z.object({
 
 export const getServerSideProps: GetServerSideProps<{
   account: Account
-  allWalletAddresses: ListWalletAddressesResponse
-  balance: FormattedAmount
+  allWalletAddresses: WalletAddressResponse[]
```
Comment (raducristianpopa) [inline packages/wallet/frontend/src/pages/account/[accountId].tsx:155]:
```suggestion
  walletAddresses: WalletAddressResponse[]
```

packages/wallet/frontend/src/pages/account/[accountId].tsx
```diff
@@ -249,26 +176,14 @@ export const getServerSideProps: GetServerSideProps<{
...
   return {
     props: {
       account: accountResponse.result,
-      allWalletAddresses: walletAddressesResponse.result,
-      balance: formatAmount({
-        value: balance.toString(),
-        assetCode: accountResponse.result.assetCode,
-        assetScale: 9
-      })
+      allWalletAddresses: walletAddressesResponse.result
```
Comment (raducristianpopa) [inline packages/wallet/frontend/src/pages/account/[accountId].tsx:186]:
```suggestion
      walletAddresses: walletAddressesResponse.result
```

raducristianpopa (commented):
Some comments:
1. Do we still want to keep the `wmTransactions` table?
1. Can we remove the `incomingBalance`, `outgoingBalance`, `isWM` columns from the `walletAddress` table?
1. Can we get rid of the environment variables like `BASE_ASSET_SCALE`, `MAX_ASSET_SCALE`, etc. since they were introduced specifically for the WM wallet addresses?

packages/wallet/backend/migrations/20240903125600_update_wallet_addresses.js
sanducb: Looking into the db, I saw that WAs have a null asset code, except for those that are WM enabled. This migration is only for correcting that and making all the WAs have their respective asset code.

raducristianpopa: Looking at the migration history is looks like it was introduced in `20231018084125_update_payment_pointers_table.js`, when the `isWM` flag was added as well. Not sure if this column is still needed?

CC @dragosp1011 @beniaminmunteanu

beniaminmunteanu: yes it makes sense for this to be added due to the introduction of the WM payment pointers. 
That would be the only case I think of that wallet addresses would have different assets than the accounts that hold those WA.

Not needed anymore IMO.

Tymmmy:
> Some comments:
> 
> 1. Do we still want to keep the `wmTransactions` table? @dragosp1011 we should remove probably
> 2. Can we remove the `incomingBalance`, `outgoingBalance`, `isWM` columns from the `walletAddress` table? @also removeable @dragosp1011 
> 3. Can we get rid of the environment variables like `BASE_ASSET_SCALE`, `MAX_ASSET_SCALE`, etc. since they were introduced specifically for the WM wallet addresses?
Base asset scale is used now so we don't have the hardcoded 2, but max asset scale can be removed - I will do it

dragosp1011 (approved):

raducristianpopa (approved):

Tymmmy (approved):

PR #1530, opened 2024-09-04
feat(wallet/frontend): replace rapyd with gatehub for KYC

### LINKED ISSUE #1529: [GateHub] - Replace Rapyd KYC with GateHub KYC
### Context

We need to get rid of Rapyd, and one of the replacements we need to do is KYC. We will be using GateHub KYC to check users.
We will not be able pre fill the fields for KYC since we are going to use an iframe that points to GateHub - the users will have to fill in all the fields.

### Todos

- [ ] remove all KYC forms and use Gatehub forms in Iframe

raducristianpopa:
### Context

Closes #1529.

### Changes
- Adds a standalone KYC page since everything will happen in an iframe
- Removes previous KYC pages (personal/identity verification)
- Updates user service: remove all obsolete API calls from the frontend
- Adds `getBearerToken` method to retrieve the user token for GateHub iframes (should be used for the other iframes as well)


packages/wallet/frontend/src/pages/kyc.tsx
```diff
@@ -0,0 +1,74 @@
...
+import AuthLayout from '@/components/layouts/AuthLayout'
+import { userService } from '@/lib/api/user'
+import { NextPageWithLayout } from '@/lib/types/app'
+import { GetServerSideProps, InferGetServerSidePropsType } from 'next/types'
+import { useEffect } from 'react'
+
+type KYCPageProps = InferGetServerSidePropsType<typeof getServerSideProps>
+
+const KYCPage: NextPageWithLayout<KYCPageProps> = ({ url }) => {
+  // const [openDialog, closeDialog] = useDialog()
```
Tymmmy, line 11: do we need the comments?

raducristianpopa: Commented out for now. We will keep/remove as needed once we can test the KYC flow.

dragosp1011:
it was inherited by https://github.com/interledger/testnet/pull/1643 and merged

Issue #1522, opened 2024-09-03
[Cards] - create template for cards and buttons


Tymmmy:
### Context

The Cards and buttons repeat themselfs over various pages in different types or buttons.
We should implement a component that can be reused for every page.

### Todos

- [ ] implement a crads image component - 3 types - normal, details, frozen
- [ ] implement the button menu actions bellow cards display

PR #1544, opened 2024-09-05
fix(wallet/frontend): use css for user carsd instead of svgs

raducristianpopa:
### Context

Continuation of: #1522.

### Changes
- [x] Create component for card front
- [ ] Create component for card back
- [ ] Construct card with CSS instead of using SVGs


raducristianpopa (commented):
Will send two additional PRs after this one gets merged:
- updating card colors - we are going to have a white card
- separate card side view from card state (frozen or not)

packages/wallet/frontend/src/components/userCards/UserCard.tsx
```diff
@@ -1,63 +1,130 @@
...
-import { DetailsCard } from '../icons/DetailsCard'
-import Image from 'next/image'
-import { CardTypes } from '@/pages/card'
+import { Chip, GateHubLogo, MasterCardLogo } from '../icons/UserCardIcons'
+import { cn } from '@/utils/helpers'
 
-type UserCardProps = {
-  type: CardTypes
+const CARD_TYPE = {
+  normal: 'normal',
```
Tymmmy, line 7: should w erename here as well the type? front - back - frozen?

PR #1546, opened 2024-09-05
feat: implement initial gatehub client

dragosp1011:
### Context

- fixes #

### DIFF
--- a/packages/wallet/backend/src/gatehub/client.ts
+++ b/packages/wallet/backend/src/gatehub/client.ts
@@ -170,10 +170,18 @@ export class GateHubClient {
   private async request<T>(
     method: HTTP_METHODS,
     url: string,
-    body?: string
+    body?: string,
+    managedUserUuid?: string
   ): Promise<T> {
     const timestamp = Date.now().toString()
-    const headers = this.getRequestHeaders(timestamp, method, url, body ?? '')
+    const headers = this.getRequestHeaders(
+      timestamp,
+      method,
+      url,
+      body ?? '',
+      managedUserUuid
+    )
+
     try {
       const res = await axios<T>({
         method,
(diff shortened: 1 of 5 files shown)


beniaminmunteanu (approved):

packages/wallet/backend/src/gatehub/client.ts
```diff
@@ -0,0 +1,239 @@
...
+  private clientIds = SANDBOX_CLIENT_IDS
+  private mainUrl = 'sandbox.gatehub.net'
+  constructor(
+    private env: Env,
+    private logger: Logger
+  ) {
+    if (this.env.NODE_ENV === 'production') {
+      this.clientIds = PRODUCTION_CLIENT_IDS
+      this.mainUrl = 'gatehub.net'
+    }
```
raducristianpopa, line 33: How are we going to make a difference between the production and the staging environment (for the deployed versions).

packages/wallet/backend/src/gatehub/client.ts
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  async getDepositUrl(): Promise<string> {
+    const token = this.getIframeAuthorizationToken(
+      this.clientIds.onOffRamp,
+      DEFAULT_APP_SCOPE
+    )
+
+    return `${this.rampUrl}/?paymentType=${PAYMENT_TYPE.deposit}&bearer=${token}`
+  }
```
raducristianpopa, line 68: Should only have one method `getRampUrl(paymentType: PAYMENT_TYPE}` instead of two separate functions? The only difference is the `paymentType` query parameter.

packages/wallet/backend/src/gatehub/client.ts
dragosp1011: good question. Should we use another env for deployed test env?
there also can be added a variable that will indicate if sandbox or production should be used for GateHub

raducristianpopa: > Should we use another env for deployed test env?
We can use `staging` as another `NODE_ENV` value. If we do not have have a lot of checks in the codebase related to the environment, this would fit best. Otherwise, we can just add another env variable for this.

packages/wallet/backend/src/gatehub/client.ts
dragosp1011: I will keep this way for now and changed if it makes sense on endpoint implementation

raducristianpopa (approved):
We can make the environment changes in a separate PR if needed.

PR #1555, opened 2024-09-06
feat(wallet/frontend): separate card state from side view

raducristianpopa:
### Context

Continuation of #1544.

### Changes
- the `frozen` property will be returned by the API, therefore the side view should be separated from the frozen state
- Switching back to card context since a mock API was added and it would help us prototype faster
- Updates card actions component
- Button layout shifting was also fixed in here; there was an issue with the icon sizes


raducristianpopa:
This should be reviewed & merged after #1554.

packages/wallet/frontend/src/lib/api/card.ts
```diff
@@ -0,0 +1,175 @@
...
+  async getDetails() {
+    return Promise.resolve({
+      success: true,
+      message: 'Mocked getDetails',
+      result: {
+        name: 'John Doe',
+        number: '4242 4242 4242 4242',
+        expiry: '01/27',
+        cvv: 321,
+        isFrozen: Math.random() > 0.5 ? true : false
```
raducristianpopa, line 121: Note to reviewer:

The `isFrozen` property is random at the moment. Therefore, freezing/unfreezing might not happen on the first click of the button - click a few times until the state changes.

Tymmmy (approved):

PR #1556, opened 2024-09-06
fix: forgotten password - extend reset token expire time

### LINKED ISSUE #1541: [BUG] Forgotten password - to short time to reset
### Steps to reproduce

1. Go to Login page
2. Click on Forgot password and fill in your email
3. You get an email with reset link, but wait at least 15 minutes, maybe you take a call...or a meeting
4. Click on reset link, it is expired

### Expected result

We should have reset password after forgotten it, for at least a half a day, a few hours.

### Actual result

After 10 minutes approx. the reset link is not working anymore

### Screenshots or videos

_No response_

### Additional context

_No response_

### Operating system

macOS

### Operating system version

_No response_

### Browsers

Chrome

### Browser version

_No response_

adrianboros:
### Context

- fixes #1541 

### Changes
- set password reset token expire after 12h

### DIFF
--- a/packages/wallet/backend/src/user/service.ts
+++ b/packages/wallet/backend/src/user/service.ts
@@ -72,7 +72,7 @@ export class UserService implements IUserService {
 
     const resetToken = getRandomToken()
     const passwordResetToken = hashToken(resetToken)
-    const passwordResetExpiresAt = new Date(Date.now() + 10 * 60 * 1000)
+    const passwordResetExpiresAt = new Date(Date.now() + 12 * 3600 * 1000)
 
     await User.query()
       .findById(user.id)
--- a/packages/wallet/backend/src/user/service.ts
+++ b/packages/wallet/backend/src/user/service.ts
@@ -72,7 +72,7 @@ export class UserService implements IUserService {
 
     const resetToken = getRandomToken()
     const passwordResetToken = hashToken(resetToken)
-    const passwordResetExpiresAt = new Date(Date.now() + 12 * 3600 * 1000)
+    const passwordResetExpiresAt = new Date(Date.now() + 8 * 3600 * 1000)
 
     await User.query()
       .findById(user.id)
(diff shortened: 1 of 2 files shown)


packages/wallet/backend/src/user/service.ts
```diff
@@ -72,7 +72,7 @@ export class UserService implements IUserService {
 
     const resetToken = getRandomToken()
     const passwordResetToken = hashToken(resetToken)
-    const passwordResetExpiresAt = new Date(Date.now() + 10 * 60 * 1000)
+    const passwordResetExpiresAt = new Date(Date.now() + 12 * 3600 * 1000)
```
raducristianpopa, line 75: Not to keen about this. I feel like having the password reset expiry at 12 hours might be a bit too much, with the fact that we are going to have a real production environment as well.

I feel like 30 minutes, 1 hour or 2 hours - one of these options, would probably fit best. 


We should also mention in the email for how long the reset link is going to be available - somewhere around here I would say:

https://github.com/interledger/testnet/blob/60eb19321277f14c0fe44819ef443d94e62fef05/packages/wallet/backend/src/email/templates/forgotPassword.ts#L30-L41

CC @Tymmmy @beniaminmunteanu @dragosp1011

Tymmmy: Feedback directly from Brianna through Tseli, we should have this around half a day, let's do 8 working hours then. But for sure not less. @raducristianpopa

Tymmmy, line 75: Set the hours to 8 hours, and update the email someweher to mention the expiry time. 
'Password reset link will expire in 8 hours.' - somewhere

Tymmmy (approved):

packages/wallet/backend/src/user/service.ts
adrianboros: updated to 8h and set notification message in the email

PR #1570, opened 2024-09-10
feat: button hover actions

### LINKED ISSUE #1569: Add light mode hover layouts for buttons with icons
### Context

Currently light mode buttons with text have hover effect as underline the text.
When the button is an icon , no text, there is no hover effect.

### Todos

- [x] add hover effect to buttons that only have an icon inside
- [x] hover effect should be a shadow

Tymmmy:
### Context

- fixes #1569 

### Changes
- added hover button layout for light mode buttons with only icon inside

### DIFF
--- a/packages/wallet/frontend/tailwind.config.js
+++ b/packages/wallet/frontend/tailwind.config.js
@@ -67,6 +67,16 @@ module.exports = {
           '0 0 0.2rem #fff',
           '0 0 1rem rgb(var(--pink-neon))',
           '0 0 0.4rem rgb(var(--pink-neon))'
+        ],
+        'glow-svg-green': [
+          '0 0 0.2rem #fff',
+          '0 0 0.2rem rgb(var(--green-dark))',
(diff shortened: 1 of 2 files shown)


packages/wallet/frontend/tailwind.config.js
```diff
@@ -67,6 +67,16 @@ module.exports = {
...
+        ],
+        'glow-svg-green': [
+          '0 0 0.2rem #fff',
+          '0 0 0.2rem rgb(var(--green-dark))',
+          '0 0 1rem rgb(var(--green-dark))'
+        ],
+        'glow-svg-orange': [
+          '0 0 0.2rem #fff',
+          '0 0 0.2rem rgb(var(--orange-dark))',
+          '0 0 1rem rgb(var(--orange-dark))'
```
raducristianpopa, line 79: We need to define the hover states for dark mode as well. At the moment we are using the green and orange colors for both themes.

raducristianpopa (approved):

PR #1586, opened 2024-09-12
fix: Resend verification email

### LINKED ISSUE #1560: [BUG] Resend verification email
### Steps to reproduce

1. Create a new user with email address and receive verification email
2. Accidentaly delete all traces of your verification email
3. Try to log in, but EMail address is not verified error appears
4. No way to verify your email

### Expected result

There should be a way to resend verification email, when something happens with the original email. - EDGE CASE

### Actual result

There is no way of verifying an email again if something happens to the original verification email.

### Screenshots or videos

image

### Additional context

ToDO - 
- [ ] Add a Resend verification link when error message appears: Email address not verified.
- [ ] Make sure verification link is available for at least one day after email sent.
- [ ] Make sure Resend verification link is working and email is sent to the user again.

### Operating system

macOS

### Operating system version

_No response_

### Browsers

Chrome

### Browser version

_No response_

adrianboros:
### Context

- fixes #1560 

### Changes
 - add support to resend account validation email

### DIFF
--- a/packages/wallet/backend/src/app.ts
+++ b/packages/wallet/backend/src/app.ts
@@ -171,6 +171,7 @@ export class App {
     router.get('/reset-password/:token/validate', userController.checkToken)
     router.post('/reset-password/:token', userController.resetPassword)
     router.post('/verify-email/:token', authController.verifyEmail)
+    router.post('/resend-verify-email', authController.resendVerifyEmail)
     router.patch('/change-password', isAuth, userController.changePassword)
 
     // Me Endpoint
--- a/packages/wallet/backend/src/auth/service.ts
+++ b/packages/wallet/backend/src/auth/service.ts
@@ -6,8 +6,11 @@ import type { UserService } from '@/user/service'
 import { getRandomToken, hashToken } from '@/utils/helpers'
 import { EmailService } from '@/email/service'
 import { Logger } from 'winston'
-import { Unauthorized } from '@shared/backend'
+import { Unauthorized, NotVerified } from '@shared/backend'
 
+interface resendVerifyEmailArgs {
+  email: string
+}
 interface AuthorizeArgs {
   email: string
   password: string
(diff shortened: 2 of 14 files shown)


packages/wallet/backend/src/app.ts
```diff
@@ -171,6 +171,7 @@ export class App {
     router.get('/reset-password/:token/validate', userController.checkToken)
     router.post('/reset-password/:token', userController.resetPassword)
     router.post('/verify-email/:token', authController.verifyEmail)
+    router.post('/resend-verify-email', authController.resendVerifyEmail)
```
Tymmmy, line 174: Token not needed here, as the other resend or reset password calls?

packages/wallet/backend/tests/auth/service.test.ts
```diff
@@ -116,6 +116,30 @@ describe('Authentication Service', (): void => {
     })
   })
 
+  describe('ResendVerifyEmail', () => {
+    it('should return undefined if verification successfully sent', async () => {
```
Tymmmy, line 120:
```suggestion
    it('should return undefined if verification email successfully sent', async () => {
```

Tymmmy (dismissed):

lengyel-arpad85 (dismissed):

Tymmmy (approved):

packages/wallet/backend/src/auth/service.ts
```diff
@@ -54,21 +57,44 @@ export class AuthService implements IAuthService {
...
+    if (!user) {
+      throw new Unauthorized('Invalid account')
+    }
+    const token = getRandomToken()
+    await this.userService.resetVerifyEmailToken({
+      email,
+      verifyEmailToken: hashToken(token)
+    })
+
+    await this.emailService.sendVerifyEmail(email, token).catch((e) => {
```
dragosp1011, line 75: because the error is caught here, the response to frontend will be that email was successfully send
is this expected?

packages/wallet/backend/src/auth/service.ts
```diff
@@ -54,21 +57,44 @@ export class AuthService implements IAuthService {
...
     // TODO: Prevent timing attacks
     if (!user) {
-      throw new Unauthorized('Invalid credentials')
+      throw new Unauthorized('Invalid credentials.')
     }
 
-    const isValid = await user.verifyPassword(args.password)
-    if (!isValid) {
-      throw new Unauthorized('Invalid credentials')
+    if (!user.isEmailVerified) {
```
dragosp1011, line 91: verifying the password first is the correct approach.
Telling the user that the email is not verified even if the password is incorrect increases security risks.

packages/wallet/frontend/src/lib/api/user.ts
```diff
@@ -139,6 +139,9 @@ export const verifyIdentitySchema = z
 export const forgotPasswordSchema = z.object({
   email: z.string().email({ message: 'Email is required' })
 })
+export const resendVerificationEmailSchema = z.object({
```
dragosp1011, line 142: move the schema to shared lib and use the same schema in both frontend and backend

dragosp1011: as forgot password has the same schema use one generic schema for both cases

packages/wallet/frontend/src/lib/api/user.ts
```diff
@@ -379,6 +393,19 @@ const createUserService = (): UserService => ({
     }
   },
 
+  async resendVerifyEmail(args) {
+    try {
+      const response = await httpClient
+        .post(`resend-verify-email/`, {
```
dragosp1011, line 399: ``resend-verify-email/`` -> `'resend-verify-email'`

Tymmmy (dismissed):

packages/wallet/backend/src/auth/service.ts
```diff
@@ -54,21 +57,44 @@ export class AuthService implements IAuthService {
...
   }
 
+  public async resendVerifyEmail({
+    email
+  }: resendVerifyEmailArgs): Promise<void> {
+    const user = await this.userService.getByEmail(email)
+
+    // TODO: Prevent timing attacks
+    if (!user) {
+      throw new Unauthorized('Invalid account')
```
dragosp1011, line 67: I recommend that we don't send an error in this case and just return a success response
user should not be able to understand if an account exists or not to reduce security risks

packages/wallet/backend/src/user/service.ts
```diff
@@ -214,6 +221,19 @@ export class UserService implements IUserService {
     })
   }
 
+  public async resetVerifyEmailToken(args: VerifyEmailArgs): Promise<void> {
+    const user = await this.getByEmail(args.email)
+
+    if (!user) {
+      throw new BadRequest('Invalid user')
```
dragosp1011, line 228: sam here, should not be an error if user does not exist

dragosp1011 (approved):

PR #1595, opened 2024-09-16
chore(docker): add new env vars

raducristianpopa:
### Context

Extracted from #1530.

### Changes
Adds new required env vars to both `dev` and `prod`.


Tymmmy (approved):
Should we remove

dragosp1011 (approved):

PR #1600, opened 2024-09-16
fix: moved chared zod schemas, updated password generation for tests

### LINKED ISSUE #1336: [Subtask] -Share types for Auth
We need to check if there are shared zod schemas and return types between frontend and backend in Test Wallet.
If there are, we should move them to the shared package.

adrianboros:
### Context

- fixes #1336 

### Changes
- moved chared zod schemas, updated password generation for tests


packages/wallet/backend/src/rafiki/validation.ts
```diff
@@ -1,15 +1,6 @@
 import { z } from 'zod'
 import { EventType, PaymentType } from './service'
 
-export const ratesSchema = z.object({
```
dragosp1011, line 4: rate endpoint is used by rafiki and not wallet frontend, it should not be moved in shared

packages/wallet/backend/src/rapyd/schemas.ts
```diff
@@ -37,15 +37,6 @@ export const walletSchema = z.object({
   })
 })
 
-export const ratesSchema = z.object({
```
dragosp1011, line 40: schemas for rapyd should not be moved

packages/wallet/backend/src/rafiki/validation.ts
adrianboros: removed from here and use this definition from rates/validation

packages/wallet/backend/src/rapyd/schemas.ts
adrianboros: removed from here and use this definition from rates/validation

dragosp1011 (approved):

dragosp1011 (approved):

PR #1632, opened 2024-09-19
feat: add rate limiters for sending emails and login

### LINKED ISSUE #1585: Prevent Brute force attacks - Support rate limiting login attempts
### Context

Currently a user can try to login with a bad password as much as he wants, which means the site is vulnerable to brute force attacks. This means users accounts can be compromised.

### Todos

Implement a rate limit feature to limit login attempts

### LINKED ISSUE #1610: Prevent Brute force attacks - Support rate limiting for "Forgot password" and "verify account email" emails
### Context

Currently you can resend email for forgotten password and verification email so It is possible to spam users' mailboxes by abusing the **Forgot Password** and **Verify Account** features. This can hurt our sender reputation, annoy users and flood the server with unnecessary traffic.

### Todos

Implement a rate limiting feature where we could set a limit on how many password reset emails and verify account emails can be generated for a single email address in a given time frame.

adrianboros:
### Context

- fixes #1585
- fixes #1610 

### Changes
 - add rate limiters for sending verify and forgot pass emails
 - add rate limiters on login attempts


adrianboros:
This rate limit implementation works as route middleware, which means we count the endpoint hit and NOT failed login attempts. We can hit the rate limit even with successful logins if we login/logout multiple times in a short amount of time.

Questions: 
- should we set these rate limits as env variables to be easier to update?
- for login rate limit, should we count only failed attempts and not all route hits? in order to count just failed login attempts, we can move login route rate limit logic from a middleware to a service and use it in the `authService.authorize` function.

lengyel-arpad85 (changes requested):
looks good, 
I would add some tests as well, since we already have some for login and resend in packages/wallet/backend/tests/auth/controller.test.ts
extend this or add some for the rate limit

lengyel-arpad85:
> This rate limit implementation works as route middleware, which means we count the endpoint hit and NOT failed login attempts. We can hit the rate limit even with successful logins if we login/logout multiple times in a short amount of time.
> 
> Questions:
> 
> * should we set these rate limits as env variables to be easier to update?
> * for login rate limit, should we count only failed attempts and not all route hits? in order to count just failed login attempts, we can move login route rate limit logic from a middleware to a service and use it in the `authService.authorize` function.

- yes I think having the rate limits in ENV variables is a good idea
- if we are trying to prevent brute force attacks on login then current approach of route middleware is ok, I saw that you allow 3 maxAttempts before the pause, so that is ok from me

docker/dev/.env.example
```diff
@@ -9,6 +9,18 @@ GATEHUB_GATEWAY_UUID=
...
+SEND_EMAIL_RATE_LIMIT_RESET_INTERVAL_IN_SECONDS=3600
+LOGIN_RATE_LIMIT=3
+LOGIN_RATE_LIMIT_PAUSE_IN_SECONDS=600
+LOGIN_RATE_LIMIT_RESET_INTERVAL_IN_SECONDS=3600
+LOGIN_IP_RATE_LIMIT=30
+LOGIN_IP_RATE_LIMIT_PAUSE_IN_SECONDS=3600
+LOGIN_IP_RATE_LIMIT_RESET_INTERVAL_IN_SECONDS=3600
+LOGIN_IP_BLOCK_RATE_LIMIT=500
+LOGIN_IP_BLOCK_RATE_LIMIT_PAUSE_IN_SECONDS=86400
+LOGIN_IP_BLOCK_RATE_LIMIT_RESET_INTERVAL_IN_SECONDS=86400
```
adrianboros, line 23: We have quite a few parameters. For each rate limiter we have or we may add later, we have 3 env params
- rate limit - number of times an resource can be used
- pause  - pause time in seconds until the resource can be used again by the same user or IP
- reset interval - reset time in seconds, is the time until we reset the counter of resource usage by the same user/IP

lengyel-arpad85 (approved):

docker/dev/.env.example
dragosp1011: there are too many env variables related to rate limiting in my opinion and there will be even more based on this structure as we add it to more routes
I think we should have just a few: enabled rate limit, some global limitation and points and maybe some coeficient based on what the rate limit variables are calculated

what do you think? @adrianboros @lengyel-arpad85 @sanducb @beniaminmunteanu

adrianboros: @dragosp1011  we have now only two optional env variables to enable rate limiting and three possible levels for limits

dragosp1011 (dismissed):

beniaminmunteanu (approved):

lengyel-arpad85 (approved):

PR #1657, opened 2024-09-30
fix(wallet): await on get balance

dragosp1011:
### Context

- fixes #

### DIFF
--- a/packages/wallet/backend/src/gatehub/types.ts
+++ b/packages/wallet/backend/src/gatehub/types.ts
@@ -95,7 +95,7 @@ export interface IWalletBalance {
 interface IVault {
   uuid: string
   name: string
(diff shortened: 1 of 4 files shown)


packages/wallet/backend/src/gatehub/types.ts
```diff
@@ -95,7 +95,7 @@ export interface IWalletBalance {
 interface IVault {
   uuid: string
   name: string
-  assetCode: string
+  asset_code: string
   createdAt: string
   updatedAt: string
```
sanducb, line 100: Shouldn't these also have snake case?

sanducb (approved):
LGTM, only left one question

sanducb (approved):

PR #1671, opened 2024-10-02
1658 gatehub add tests for gatehub module

### LINKED ISSUE #1658: [GateHub]  - Add tests for gatehub module
### Context

For the gatehub integration module, not all sections are covered by tests. 
Add test for gatehub module to:

### Todos

- [x]  Cover controller actions scenarios
- [x]  Cover service actions scenarios

adrianboros:
### Context

- fixes #1658

Added tests for gatehub module in wallet backend

### DIFF
--- a/packages/wallet/backend/tests/gatehub/controller.test.ts
+++ b/packages/wallet/backend/tests/gatehub/controller.test.ts
@@ -1,0 +1,161 @@
+import { env } from '@/config/env'
+import { Cradle, createContainer } from '@/createContainer'
+import { AuthService } from '@/auth/service'
+import { NextFunction, Request, Response } from 'express'
+import { createApp, TestApp } from '@/tests/app'
+import {
+  MockRequest,
+  MockResponse,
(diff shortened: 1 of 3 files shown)


packages/wallet/backend/src/rafiki/service.ts
```diff
@@ -126,7 +234,8 @@ export class RafikiService implements IRafikiService {
     if (
       [
         EventType.OutgoingPaymentCreated,
-        EventType.OutgoingPaymentCompleted
+        EventType.OutgoingPaymentCompleted,
+        EventType.OutgoingPaymentFailed
```
adrianboros, line 172: Bug fix here: We have to parse `debitAmount` even for  `OutgoingPaymentFailed` event otherwise it throws a bad request  error - few lines below - and the event is not handled at all

dragosp1011 (approved):

dragosp1011 (approved):

PR #1675, opened 2024-10-03
feat(wallet/backend): add script for managed user and customer creation

### LINKED ISSUE #1667: [Cards] Wallet Backend - Customer creation
### Context

Implement script to create card customers that will be associated to cards.

The script should execute the following calls:

1. Create managed users on GateHub
2. Get application products to get the product code for both account and card.
3. Create customer with the above codes.

### Todos

Implement and execute the script for the sandbox environment.

sanducb:
### Context

- fixes #1667 

### Changes

This PR adds the script for managed user and customer creation as well as the updated types.

### Note: Still WIP


raducristianpopa (commented):
Based on a discussion with Bogdan:

- when we create the managed user we want to add the payment pointer in the `meta` object.
  - if the user signs up and verifies the email, we can automatically create the payment pointer on our side with the correct value (i.e: 059)
  - waiting for response ^
- whitelisting: whenever someone tries to sign up, we fetch all managed users and we check if the email exists
  - if the user exists, we can retrieve the payment pointer from the `meta` object and create the payment pointer
  - if the user doesn't exist: _TBD_

packages/wallet/backend/migrations/20241008143459_add_customer_id_to_users.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.table('users', function (table) {
+    table.string('customerId').unique().nullable()
```
raducristianpopa, line 7: I don't think this should be nullable?

CC @sanducb

packages/wallet/backend/src/gatehub/client.ts
```diff
@@ -178,6 +179,22 @@ export class GateHubClient {
...
+  /**
+   * The meta was createad as `meta.meta.[property]`
+   * We should be aware of this when the user signs up (for production)
+   */
+  async updateMetaForManagedUser(
+    userUuid: string,
+    meta: Record<string, string>
+  ): Promise<void> {
+    const url = `${this.apiUrl}/auth/v1/users/managed`
+    const body = { meta }
```
raducristianpopa, line 194: This was the reason for the faulty `meta` object.

packages/wallet/backend/src/user/model.ts
```diff
@@ -20,6 +20,7 @@ export class User extends BaseModel {
 
   public kycVerified!: boolean
   public gateHubUserId?: string
+  public customerId?: string
```
raducristianpopa, line 23: Same here. We should have the `customerId` at all times.

packages/wallet/backend/src/gatehub/client.ts
raducristianpopa: Added a comment and updated the types for the `meta` object here: https://github.com/interledger/testnet/pull/1675/commits/930e7509ebc3a98425963fc6a6ef980059a27d87.

packages/wallet/backend/src/cardUsers.ts
```diff
@@ -0,0 +1,110 @@
...
+import { ICreateCustomerRequest } from '@/card/types'
+
+interface UserData {
+  email: string
+  firstName: string
+  lastName: string
+  ppNumber: string
+}
+
+const entries = `John;Doe;john@doe.com:8888
```
dragosp1011, line 12: is this for example porposes? can the list be read from an external file?

packages/wallet/backend/src/cardUsers.ts
```diff
@@ -0,0 +1,110 @@
...
+      logger.info(`Created managed user for ${email}: ${managedUser.id}`)
+
+      await gateHubClient.connectUserToGateway(managedUser.id, GATEWAY_UUID)
+
+      logger.info(
+        `Connected user ${managedUser.id} - ${managedUser.email} to gateway`
+      )
+
+      const user = await gateHubClient.getWalletForUser(managedUser.id)
+      const walletAddress = user.wallets[0].address
```
dragosp1011, line 58: should we use the default wallet created instead of creating on for each account?

dragosp1011 (approved):

packages/wallet/backend/src/cardUsers.ts
raducristianpopa: This is how the entries should be added. It can be updated to read from an external file as well, but for simplicity reason we decided to just put the CSV contents directly in the variable.

packages/wallet/backend/src/cardUsers.ts
raducristianpopa: `getWalletForUser` retrieves the already created account - when the managed user is created. We are not creating another wallet here.

raducristianpopa:
> I think we can already rename GATEHUB_NAME_ON_CARD_PREFIX to something like GATEHUB_NAME_ON_CARD
Oh, yes! A lot better than keeping the card prefix there.

dragosp1011 (approved):

PR #1692, opened 2024-10-07
feat(workflows): separate image for boutique-frontend for jopacc

beniaminmunteanu:
### Changes

Github action for build&publish now also takes into account the need of having a different image created and published for 
jopacc, for the boutique-frontend application.

This is due to the fact that vite uses build-time environment variables, and as such, different env vars require different builds


.github/workflows/build-publish.yaml
```diff
@@ -80,6 +89,8 @@ jobs:
           context: ./
           push: true
           file: ${{ matrix.package.path }}/Dockerfile.prod
+          build-args: |
+            ${{ matrix.package.identifier == 'boutique:frontend' && matrix.package.env_var && 'VITE_API_BASE_URL=https://api-boutique.jopacc.openpayments.directory' || '' }}
```
raducristianpopa, line 93: We will have to specify the URL for the original Rafiki Boutique as well and also push them to different tags. Since this will build twice it will be overwrite the previous built image with `env_var: false` because both of them are pushed to the `latest` tag.

beniaminmunteanu: I think they will be pushed in separate tags as the tag contains `matrix.package.name` in it, and in the matrix packages array, the items have different names: 

- name: rafiki-boutique-frontend
- name: rafiki-boutique-frontend-jopacc

And regarding the url for the original rafiki boutique, it was not set to anything before, and as such the default value is set from codebase: localhost:3004. 
You want me to add localhost:3004 as url in here as well, and not rely on the default from the codebase?

raducristianpopa: Got confused by the multiple matrices.

You are right, but this means that we can get rid of the extra build here, since it's just testing the build process:

https://github.com/interledger/testnet/blob/784efeb08f6de6d03f8eaa06d6d385af5f768292/.github/workflows/build-publish.yaml#L13-L26

> You want me to add localhost:3004 as base_api_url for the original Rafiki Boutique as well, and not rely on the default from the codebase?

We will have to specify the actual API url, not `localhost:3004`, just like we do for the JoPACC one (in our current status it will have to be `https://api.rafiki.boutique`.)

beniaminmunteanu: Oh so the artefacts of the build step that runs first are not used in the docker image build that is done as a step of the publish job? It's just so that it fails early?

So changes to the build job can be entirely reverted, the env variable added there makes no difference, all that is needed is the one done later, added as build-args to the docker build command?

.github/workflows/build-publish.yaml
```diff
@@ -17,12 +17,13 @@ jobs:
...
-            'wallet:frontend',
-            'boutique:backend',
-            'boutique:frontend'
-          ]
+        [
+          'wallet:backend',
+          'wallet:frontend',
+          'boutique:backend',
+          'boutique:frontend'
+        ]
```
raducristianpopa: Make sure the spacing is correct. There is still a diff for these.

.github/workflows/build-publish.yaml
```diff
@@ -33,13 +34,13 @@ jobs:
         with:
           node-version: '20'
           cache: 'pnpm'
-      - name: Install dependencies for ${{ matrix.package }}
+      - name: Install dependencies for ${{ matrix.package.name }}
```
raducristianpopa, line 37: Can be reverted. to `matrix.package`.

.github/workflows/build-publish.yaml
```diff
@@ -33,13 +34,13 @@ jobs:
...
         shell: bash
         run: |
-          pnpm ${{ matrix.package }} install --frozen-lockfile
+          pnpm ${{ matrix.package.name }} install --frozen-lockfile
```
raducristianpopa, line 40: Ditto above.

.github/workflows/build-publish.yaml
```diff
@@ -33,13 +34,13 @@ jobs:
...
       - name: Run build
         run: |
-          pnpm ${{ matrix.package }} build
+          pnpm ${{ matrix.package.name }} build
```
raducristianpopa, line 43: Ditto above.

Tymmmy (dismissed):

.github/workflows/build-publish.yaml
```diff
@@ -80,6 +87,9 @@ jobs:
           context: ./
           push: true
           file: ${{ matrix.package.path }}/Dockerfile.prod
+          build-args: |
+            ${{ matrix.package.identifier == 'boutique:frontend' && format('VITE_API_BASE_URL={0}', matrix.package.api_base_url) }}
+            ${{ matrix.package.identifier == 'boutique:frontend' && format('VITE_CURRENCY={0}', matrix.package.currency) }}
```
raducristianpopa, line 92: This results in:
```
/usr/bin/docker buildx build --build-arg false --build-arg false 
```

Note the `--build-arg false`. Add `format('VITE_API_BASE_URL={0}', matrix.package.api_base_url) || ''` again. This way we are not going to pass `false` multiple times as a build argument.

Workflow run when using ` || ''`:  https://github.com/interledger/testnet/actions/runs/11217123924/job/31178030911
Workflow run when not using ` || ''`: https://github.com/interledger/testnet/actions/runs/11218519020/job/31182550051

beniaminmunteanu: good catch, thanks!

raducristianpopa (approved):

PR #1697, opened 2024-10-07
feat(wallet/backend): add customer creation on login for sandbox environment

### LINKED ISSUE #1714: [Cards] Wallet Backend - Add customer creation on login for sandbox environment
### Context

We should be able to locally test `cards` and `customer` APIs on the sandbox environment without running scripts manually to create the needed resources i.e. cards and customers. Therefore, customers and cards should be created at login.

### Todos

Implement logic to create customer on login for sandbox environment

sanducb:
### Context

- fixes #1714 

### Changes


packages/wallet/backend/src/gatehub/service.ts
```diff
@@ -69,9 +73,62 @@ export class GateHubService {
...
+    }
+    const productCode = (await cardProducts).map((product) => product.code)
+
+    await this.gateHubClient.createCustomer({
+      walletAddress: account.gateHubWalletId,
+      account: {
+        productCode: productCode[0],
+        card: {
+          productCode: productCode[0],
+          nameOnCard: `INTERLEDGER $ILP.DEV/${walletAddressName}`
```
sanducb, line 122: Should we leave this hardcoded for now as we now the domain will be ilp.dev for sandbox?

packages/wallet/backend/src/gatehub/types.ts
```diff
@@ -47,6 +47,7 @@ export interface IGetUserStateResponse {
     address_street1: string
     address_street2: string
   }
+  meta: Record<string, string>
```
sanducb, line 50: We should ask to receive this field on fetching the user state if we don't already (cannot find it in the docs though)

raducristianpopa: There is no `GET` options for a user?

packages/wallet/backend/src/gatehub/service.ts
raducristianpopa: I would put in in an environment variable as well. This way you can reuse it in 
- https://github.com/interledger/testnet/pull/1675

since you are still creating the container there as wel..

packages/wallet/backend/src/account/service.ts
```diff
@@ -70,18 +70,25 @@ export class AccountService implements IAccountService {
...
 
-    const result = await this.gateHubClient.createWallet(
-      user.gateHubUserId,
-      args.name
-    )
+    const result = await this.gateHubClient.getWalletForUser(user.gateHubUserId)
+    // The user should not have more than one wallet address anyway
+    // Only on production GateHub wallet address would already exist
+    let gateHubWalletId = result.wallets[0].address
+    if (!gateHubWalletId) {
```
beniaminmunteanu, line 77: discussed to be changed to check if execution of function is for default card-holding account in production, where users are expected to have the gatehub wallet account already, instead of checking if wallet already exists 

as this would disrupt flow for users just creating new accounts

beniaminmunteanu (approved):

packages/wallet/backend/src/gatehub/service.ts
```diff
@@ -69,7 +79,80 @@ export class GateHubService {
...
+    const customerId = customer.customers.id!
+
+    await User.query().findById(userId).patch({
+      customerId
+    })
+
+    await this.gateHubClient.updateMetaForManagedUser(managedUserId, {
+      paymentPointer: paymentPointer.url,
+      customerId
+    })
```
raducristianpopa, line 153: Do we still need to update the meta in this case?

packages/wallet/backend/src/account/service.ts
```diff
@@ -70,18 +71,27 @@ export class AccountService implements IAccountService {
...
-    const result = await this.gateHubClient.createWallet(
-      user.gateHubUserId,
-      args.name
-    )
+    let gateHubWalletId
+    if (args.isDefaultCardsAccount) {
+      const result = await this.gateHubClient.getWalletForUser(
+        user.gateHubUserId
+      )
+      gateHubWalletId = result.wallets[0].address
```
raducristianpopa, line 79: Just for my clarity: With the `isDefaultCardsAccount`: if the user already has a managed user created, we fetch the wallet, otherwise the create one?

packages/wallet/backend/src/gatehub/service.ts
sanducb: I also cannot think of a use case in the future where we would need these fields on the sandbox environment.

raducristianpopa: Yeah, lets remove it for now.

packages/wallet/backend/src/account/service.ts
sanducb: If the user already has a managed user created **and** the call to create account has `isDefaultCardsAccount` set to `true`, then we would use the already existing GateHub wallet of the managed user instead of creating a new one. Otherwise, if `isDefaultCardsAccount` is false (which happens in all cases unless specified true), then it means that some user just wants to create an account that is not related to any card so we would just create a new GateHub wallet as well.

sanducb:
As discussed with @raducristianpopa, we're now mapping the `cardId` to the `account` instead of wallet address since for now it's not important for us to have a 1:1 mapping between wallet address and card. The latest commit contains these changes.

PR #1718, opened 2024-10-13
feat(wallet/backend): lock cards API for staging

raducristianpopa:
### Changes
Since cards are not going to be available on staging, a middleware was added to ensure that the cards API are only available in production.


packages/wallet/backend/src/app.ts
```diff
@@ -386,4 +418,21 @@ export class App {
   async processResources() {
     process.nextTick(() => this.processPendingTransactions())
   }
+
+  ensureGateHubProductionEnv = async (
```
raducristianpopa, line 422: The middleware was not added in the `middleware` folder since I needed access to the container.

sanducb (approved):

dragosp1011 (approved):

PR #1734, opened 2024-10-15
feat: add support for GBP and ZAR in sandbox

dragosp1011:
### Context

- fixes #


Tymmmy (approved):

packages/wallet/backend/src/gatehub/client.ts
```diff
@@ -747,13 +753,12 @@ export class GateHubClient {
...
-      case 'USD':
-        return this.env.GATEHUB_VAULT_UUID_USD
-      case 'EUR':
-        return this.env.GATEHUB_VAULT_UUID_EUR
-      default:
-        throw new BadRequest(`Unsupported asset code ${assetCode}`)
+    const vaultId: string | undefined = this.vaultIds[assetCode]
+
+    if (!vaultId) {
+      throw new BadRequest(`Unsupported asset code ${assetCode}`)
```
beniaminmunteanu, line 759: might not have context of why throwing is opted for here, just making sure return undefined is also considered

beniaminmunteanu (approved):

sanducb (approved):

packages/wallet/backend/src/gatehub/client.ts
dragosp1011: there should not be any operations with vaults that we don't support. This throw should not be executed in prod, unless the application is misconfigured (unsupported assets are added in rafiki admin)

PR #1750, opened 2024-10-17
fix(wallet/backend) - create default account on staging

sanducb:
### Context

- fixes #

### DIFF
--- a/packages/wallet/backend/src/gatehub/service.ts
+++ b/packages/wallet/backend/src/gatehub/service.ts
@@ -18,8 +18,8 @@ export class GateHubService {
     private gateHubClient: GateHubClient,
     private logger: Logger,
     private env: Env,
-    private accountService?: AccountService,
-    private walletAddressService?: WalletAddressService
+    private accountService: AccountService,
+    private walletAddressService: WalletAddressService
   ) {}
 
   async getIframeUrl(iframeType: IFRAME_TYPE, userId: string): Promise<string> {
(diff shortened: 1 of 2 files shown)


beniaminmunteanu (approved):

packages/wallet/backend/src/gatehub/service.ts
```diff
@@ -99,11 +99,30 @@ export class GateHubService {
...
     }
 
     return { isUserApproved, customerId }
   }
 
+  private async createDefaultAccountForManagedUser(
+    userId: string,
+    isDefaultCardsAccount?: boolean
+  ): Promise<Account> {
+    const account = await this.accountService.createDefaultAccount(
```
raducristianpopa, line 114: Should we create a payment pointer as well?

beniaminmunteanu:
`setupSandboxCustomer` 
still checks existance of accountService and walletAddressService, which are no longer optional through this PR

packages/wallet/backend/src/gatehub/service.ts
```diff
@@ -18,8 +18,8 @@ export class GateHubService {
     private gateHubClient: GateHubClient,
     private logger: Logger,
     private env: Env,
-    private accountService?: AccountService,
-    private walletAddressService?: WalletAddressService
+    private accountService: AccountService,
```
beniaminmunteanu, line 21: `gatehubService.setupSandboxCustomer`
still checks existance of accountService and walletAddressService, which are no longer optional

raducristianpopa (approved):

PR #1752, opened 2024-10-19
feat(wallet-backend): handle manual approve flow

beniaminmunteanu:
store gatehub userState even if user was in manual approve flow
Include logs for user connected to gateway

### DIFF
--- a/packages/wallet/backend/src/gatehub/service.ts
+++ b/packages/wallet/backend/src/gatehub/service.ts
@@ -51,10 +51,8 @@ export class GateHubService {
           await this.addUserToGateway(gateHubUserId, true)
         }
 
-        const userId = await this.markUserAsVerified(gateHubUserId)
-        this.logger.info(
-          `USER ${userId} with gatehub id ${gateHubUserId} VERIFIED`
-        )
+        await this.markUserAsVerified(gateHubUserId)
+      
         break
     }
   }
@@ -231,7 +229,7 @@ export class GateHubService {
     return customerId
   }
 
-  private async markUserAsVerified(uuid: string): Promise<string> {
+  private async markUserAsVerified(uuid: string): Promise<void> {
     const user = await User.query().findOne({ gateHubUserId: uuid })
 
     if (!user) {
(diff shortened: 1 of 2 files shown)


sanducb (dismissed):

sanducb (approved):

packages/wallet/backend/src/gatehub/service.ts
```diff
@@ -40,22 +40,41 @@ export class GateHubService {
...
       case 'id.verification.accepted':
-        await this.markUserAsVerified(data.user_uuid)
+        const { user_uuid: gateHubUserId } = data
+
+        const user = await User.query().findOne({ gateHubUserId })
+
+        // if user is already verified (for manual verify cases)
+        // we skip the approveUserToGateway and overrideRiskLevel in addUserToGateway
+        // but still execute the function in order to store gatehub userState
+        if (user?.kycVerified) {
```
dragosp1011, line 50: if it is not verified we need to extract user data
if it is verified already indb, it should have the data in place

beniaminmunteanu: if it is not already verified (manually), PR keeps the old flow.

It only adds the case in which user was already verified, so that it stores the data in this case as well.

beniaminmunteanu: If we have issues with iframe sending messages in frontend, we will not receive the frontend call to `addUserToGateway`

However, we would manually approve it, and then receive the webhook from gatehub to verify the user.
Only in this case do we need to refetch userState from gatehub in order to sync our db.

Am I missing something?

raducristianpopa: I feel like this call should happen if the user KYC status is false and it doesn't have a `firstName` and `lastName` in our database.

beniaminmunteanu: agreed. 
Misinterpreted order of execution between webhook and addUserToGateway controller.

Added lastName condition

PR #1775, opened 2024-10-25
feat: fund card account after is verified

dragosp1011:
### Context

- fixes #

### DIFF
--- a/packages/wallet/backend/migrations/20241025020455_add_is_funded_to_accounts.js
+++ b/packages/wallet/backend/migrations/20241025020455_add_is_funded_to_accounts.js
@@ -1,0 +1,19 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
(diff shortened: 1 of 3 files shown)


packages/wallet/backend/src/card/service.ts
```diff
@@ -131,11 +135,13 @@ export class CardService {
...
     cypher: string
   ): Promise<void> {
-    await this.ensureAccountExists(userId, cardId)
+    const cardAccount = await this.ensureAccountExists(userId, cardId)
 
     await this.gateHubClient.changePin(token, cypher)
 
     await User.query().findById(userId).patch({ isPinSet: true })
+
+    await this.fundAccount(cardAccount)
```
raducristianpopa, line 144: This should be done at sign up instead of doing it when we fetch the cards.

dragosp1011: it is on set pin

PR #1882, opened 2025-01-09
fix: PP for cards did not display correctly everywhere

### LINKED ISSUE #1851: [BUG] Cards drop down PPs
### Steps to reproduce

1. Go to interledger.cards
2. Click on Sned or Request or Transactions
3. Click on any drop down that has PPs

### Expected result

We should see the PP name as ilp.dev/NUMBER

### Actual result

We always see ilp.interledger.cards/NUMBER

### Screenshots or videos

image

### Additional context

_No response_

### Operating system

Windows

### Operating system version

_No response_

### Browsers

Chrome

### Browser version

_No response_

Tymmmy:
### Context

- fixes #1851 

### Changes
- For cards PP, always display ilp.dev as domain


adrianboros (approved):

packages/wallet/frontend/src/components/WalletAddressesTable.tsx
```diff
@@ -101,7 +102,7 @@ export const CopyWalletAddress = () => {
   const { walletAddress } = useContext(WalletAddressRowContext)
   const [isCopied, setIsCopied] = useState(false)
 
-  let pointer = walletAddress.url
+  let pointer = replaceWalletAddressProtocol(walletAddress.url)
```
raducristianpopa, line 105: Here we are already checking if it is the card's wallet address, a few lines below. Remove one of them.

raducristianpopa (approved):

PR #1908, opened 2025-01-29
feat(transaction): add backfill trx details script

dragosp1011:
### Context

- fixes #

### DIFF
--- a/packages/wallet/backend/src/backfillTrxDetails.ts
+++ b/packages/wallet/backend/src/backfillTrxDetails.ts
@@ -1,0 +1,124 @@
+import { createContainer } from '@/createContainer'
+import { env } from '@/config/env'
+import { Account } from '@/account/model'
+import { Transaction } from '@/transaction/model'
+import { Model } from 'objection'
+import { transformBalance, urlToPaymentId } from '@/utils/helpers'
+
+async function backfillTrxDetails() {
+  const container = await createContainer(env)
+  const knex = container.resolve('knex')
+  Model.knex(knex)
+  const rafikiClient = container.resolve('rafikiClient')
+  const rafikiService = container.resolve('rafikiService')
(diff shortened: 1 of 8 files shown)


Tymmmy (approved):

packages/wallet/backend/src/backfillTrxDetails.ts
```diff
@@ -0,0 +1,124 @@
...
+import { Account } from '@/account/model'
+import { Transaction } from '@/transaction/model'
+import { Model } from 'objection'
+import { transformBalance, urlToPaymentId } from '@/utils/helpers'
+
+async function backfillTrxDetails() {
+  const container = await createContainer(env)
+  const knex = container.resolve('knex')
+  Model.knex(knex)
+  const rafikiClient = container.resolve('rafikiClient')
```
beniaminmunteanu, line 12: does resolving the injected logger not work in this case? Or did you just choose to go with plain console in this case

beniaminmunteanu: nvm i guess this is a separate running process 👍

beniaminmunteanu (approved):

packages/wallet/backend/src/backfillTrxDetails.ts
```diff
@@ -0,0 +1,124 @@
...
+    } catch (e) {
+      console.warn('Error on getting receiver wallet address', e)
+    }
+  }
+
+  async function getIncomingPaymentSecondParty(paymentId: string) {
+    return rafikiService.getIncomingPaymentSenders(paymentId)
+  }
+
+  async function backfillCardTrx(account: Account) {
```
raducristianpopa, line 33: We probably need try-catch in this function when, otherwise if one of the backfill attempts fails the process will stop.

PR #1983, opened 2025-04-11
feat(stripe): stripe controller uses gatehubWalletId

beniaminmunteanu:
moved `getGateHubWalletAddress` into gatehubService as it is now used both in rafikiService and in StripeService.

Stripe service now uses the gatehub transaction method with proper gatehubWalletId

### DIFF
--- a/packages/wallet/backend/src/gatehub/service.ts
+++ b/packages/wallet/backend/src/gatehub/service.ts
@@ -1,7 +1,7 @@
 import { GateHubClient } from '@/gatehub/client'
 import { IFRAME_TYPE } from '@wallet/shared/src'
 import { User } from '@/user/model'
-import { NotFound } from '@shared/backend'
+import { BadRequest, NotFound } from '@shared/backend'
 import {
   IAddUserToGatewayResponse,
   ICardTransactionWebhookData,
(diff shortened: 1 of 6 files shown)


packages/wallet/backend/src/gatehub/service.ts
```diff
@@ -420,4 +420,22 @@ export class GateHubService {
   private async updateUserFlag(userId: string, changes: Partial<User>) {
     await User.query().findById(userId).patch(changes)
   }
+
+  public async getGateHubWalletAddress(walletAddress: WalletAddress) {
```
dragosp1011, line 424: as the info is extracted from account model, I think the method will be better suited in account service

dragosp1011 (dismissed):

sanducb (dismissed):
I agree with @dragosp1011 on moving the method in account service, other than that looks good!

packages/wallet/backend/src/gatehub/service.ts
beniaminmunteanu: right. Will do this 👍

sanducb (dismissed):

dragosp1011 (dismissed):

dragosp1011 (dismissed):

dragosp1011 (approved):

PR #1987, opened 2025-04-25
fix(stripe): paymentIntent amount is now scale 2

beniaminmunteanu:


packages/wallet/backend/src/utils/helpers.ts
```diff
@@ -24,6 +24,13 @@ export const transformBalance = (value: number, scale: number): bigint => {
   return BigInt(Math.floor(value * 10 ** scale))
 }
 
+export const applyScale = (value: number, scale: number): number => {
```
sanducb, line 27: I think it would be better to have `scale` as an optional parameter here and default scale used will be 2.

packages/wallet/backend/src/stripe-integration/service.ts
```diff
@@ -49,7 +49,8 @@ export class StripeService implements IStripeService {
     const metadata = paymentIntent.metadata
     const receiving_address: string = metadata.receiving_address
     const currency: string = paymentIntent.currency
-    const amount: number = paymentIntent.amount
+
+    const scaledAmount = applyScale(paymentIntent.amount, 2)
```
sanducb, line 53: I think we already have a constant that defines the default asset scale so that we don't hardcode values anywhere.

sanducb (approved):

PR #2015, opened 2025-07-15
fix: quote expiration problem

### LINKED ISSUE #2014: [BUG] Fix quote expiration time
### Steps to reproduce

1. Go to a send page and make a quote
2. Wait around 30 second before pressing the Accept button on the quote
3. Press the accept on the quote
4. Internal server error appears

### Expected result

We should not have a quote expiration limit. 

### Actual result

We still have an internal server error showing when we wait to long to accep the quote.

### Screenshots or videos

_No response_

### Additional context

_No response_

### Operating system

macOS

### Operating system version

_No response_

### Browsers

Chrome

### Browser version

_No response_

Tymmmy:
### Context

- fixes #2014 

### Changes
We decided to put 'receive' as default sending type, this way, if someone does a lecture or presentation, quote will not expire

### DIFF
--- a/packages/wallet/backend/src/quote/service.ts
+++ b/packages/wallet/backend/src/quote/service.ts
@@ -131,7 +131,9 @@ export class QuoteService implements IQuoteService {
         },
         walletAddressUrl: params.receiver,
         description: params.description,
-        expiresAt: new Date(Date.now() + 1000 * 15)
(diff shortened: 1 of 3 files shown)


lengyel-arpad85 (approved):
LGTM

PR #2022, opened 2025-07-16
feat: Update registration option for cards

### LINKED ISSUE #2021: Remove registration if it is cards prod
Let's remove registration for the interledger.cards.
Request from interledger Wallet team

Tymmmy:
### Context

- fixes #2021

### DIFF
--- a/packages/wallet/frontend/src/pages/auth/login.tsx
+++ b/packages/wallet/frontend/src/pages/auth/login.tsx
@@ -16,7 +16,7 @@ import { ErrorDialog } from '@/components/dialogs/ErrorDialog'
 import { NextPageWithLayout } from '@/lib/types/app'
 import { useEffect, useState } from 'react'
 import { loginSchema } from '@wallet/shared'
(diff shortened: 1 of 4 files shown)


packages/wallet/frontend/src/pages/auth/login.tsx
```diff
@@ -158,12 +158,14 @@ const LoginPage: NextPageWithLayout = () => {
...
-          </Link>
-        </p>
+        {!FEATURES_ENABLED ? (
+          <p className="mt-auto text-center font-extralight text-green dark:text-green-neon">
+            Not a customer?{' '}
+            <Link href="signup" className="font-medium underline">
+              Create an account
+            </Link>
+          </p>
+        ) : null}
```
raducristianpopa, line 168: The whole registration process should be blocked. Even if we do not show the link, users can still navigate to the page. On the other hand, the backend route for sign up should be blocked as well, since it can be called via cURL, Postman, etc...

raducristianpopa (approved):

PR #2065, opened 2025-07-23
feat(wallet): forward erroring webhooks to wallet

RaulBR:
### Context
In certain scenarios, both Wallet and Testnet may share the same Provider account.
This implementation routes incoming requests to the appropriate Wallet application based on an environment variable and a NotFound Error.

All forwarded requests are signed and authenticated using the same secret key as the Testnet implementation.

- fixes #
[Forward Provider webhook to Wallet](https://linear.app/interledger/issue/WAL-191/gatehub-webhook-for-cards-and-wallet)


packages/wallet/backend/src/gatehub/controller.ts
```diff
@@ -74,10 +81,41 @@ export class GateHubController implements IGateHubController {
...
 
       await this.gateHubService.handleWebhook(req.body)
-
       res.status(200).json()
     } catch (e) {
+      //==========================
+      // ONLY TEMPORARY - added to forward webhooks to the wallet
+      // wallet and cards share the same Gatehub account witch in turn means
+      // they share the same webhook -> if no user is found in cards send to wallet
+      await this.sendMessageToWallet(e, req, next)
```
dragosp1011, line 90: if the response from wallet is successful, the response status should be succes so the webhook is not retried by GateHub

packages/wallet/backend/src/gatehub/controller.ts
```diff
@@ -74,10 +81,41 @@ export class GateHubController implements IGateHubController {
...
+        await axios.post(url, req.body, {
+          headers: {
+            'x-gh-webhook-signature': req.get('x-gh-webhook-signature'),
+            'x-gh-webhook-timestamp': req.get('x-gh-webhook-timestamp')
+          }
+        })
+
+        this.logger.info('GateHub webhook forwarded to wallet')
+        next(e)
+      } catch (e) {
```
dragosp1011, line 116: same name as `e` from params, can cause confusion

dragosp1011 (approved):

PR #2152, opened 2025-10-20
feat: display all grant accesses in interaction page

### LINKED ISSUE #1989: [FEATURE REQUEST]  - Grant Interaction page should display all grant access requested
### Describe the feature you would like to request

When requesting a grant, one might request multiple accesses:

Image

### Describe the solution you would like

- make sure all the needed information is also listed in the grant interaction page.

Image

### Additional context

If you have a grant, and check the response for the grant details request, you will see that the 'access' is an array, so we need to display on the grant interaction page all responses.

_No response_

ARYPROGRAMMER:
Fixes #1989 : 

Grant Interaction page now displays all requested grant accesses instead of just one.

- Removed filtering to only 'outgoing-payment' access
- Updated pending and processed grant UIs to list all accesses with full details
- Maintains existing styling and data formatting

for review: @Tymmmy and other maintainers

### DIFF
--- a/packages/wallet/frontend/src/pages/grant-interactions/index.tsx
+++ b/packages/wallet/frontend/src/pages/grant-interactions/index.tsx
@@ -69,51 +69,33 @@ const GrantInteractionPage = ({
         <div className="mt-20 text-base">
           <div>
             <span className="font-semibold">{client}</span> is requesting access
-            to your account with the following permissions:
+            to make{' '}
+            {grant.access.length > 1 ? 'a list of payments' : 'a payment'} on
+            your behalf.
           </div>
-          <div className="mt-4">
-            <div className="font-semibold mb-2">Requested Accesses:</div>
-            {grant.access.map((accessItem, index) => (
-              <div key={index} className="border border-gray-300 dark:border-gray-600 rounded p-2 mb-2">
-                <div>
-                  <span className="font-medium">Type: </span>
-                  <span>{accessItem.type}</span>
-                </div>
-                {accessItem.identifier && (
-                  <div>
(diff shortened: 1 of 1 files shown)


Tymmmy (changes requested):
AFter reviewing the issue closer, I would like to make some changes, to make sure it is not that technical for users that are not into the tech side.
Currently it looks like this with your changes:
image
When this idp page appears, it is to make a withdraw, a payment for sure, and that is what outgoing-payment really means. But not everyone understands. So let's make the following changes: when the grant.access array has only one item, leave the text and everything the way it was, with no extra access information. When it is an array, then let's display only the debited amounts list. The text should be 'is requesting access to make a list of payments on your behalf.' So here '**list**' will be the extra text when it is more then one item in grant.access. And then list the debit amounts. Same would go for the previously processed text, add 'a list of payments', and the list of amounts.

Thank you

packages/wallet/frontend/src/pages/grant-interactions/index.tsx
```diff
@@ -70,18 +69,33 @@ const GrantInteractionPage = ({
...
-              {access?.limits?.debitAmount?.formattedAmount}
-            </span>
+            to make{' '}
+            {grant.access.length > 1 ? 'a list of payments' : 'a payment'} on
+            your behalf.
           </div>
+          {grant.access.length === 1 ? (
+            <>
+              <div>
+                Wallet Address client:{' '}
```
Tymmmy: Wallet address client can be displayed for both when access is one item or more than one items.

Tymmmy (approved):
Perfect, thank you

PR #2153, opened 2025-10-21
feat: 2143 feature request frontend part of sepa   test wallet

### LINKED ISSUE #2143: [FEATURE REQUEST] Frontend part of SEPA - Test Wallet
Context

GateHub is working on a functionality to tie Payment pointer to a bank account. Meaning, if user will receive money on the specific payment pointer, that funds will be redirected to appropriate bank account through SEPA rail.

From 9th of October 2025, there is new regulation called Verification of Payee. Basically, it means that for every SEPA transaction, user must enter First name, Last name and the IBAN of the recipient. After that, this data is checked and user gets back a response if this is a match, close match, no match or other. After that user can initiate SEPA transaction or not. The most important thing here is that user must be informed about the result.

GateHub is proposing a flow in interledger.cards and in wallet.interledger-test.dev as well, because they need to test this somewhere :)

Example of payment pointer that is tied to bank account: `$ilp.gatehub.net/iban/SI56263300012039086

Form for entering First and last name should be shown only in case it is this kind of payment pointer.

Structure will be like this: /iban/IBAN_NUMBER

Flow would look like this (on the Send page)
Immediately after all the required field for verification of payee are filled, call GateHub endpoint to initiate Verification of Payee, because this request can take up to 20s (20s can be set as a timeout).

In response, you will receive nonce. This nonce needs to be passed when incoming payment will be created. IT should be passed inside metadata field as a value of a key named vopNonce

Todos

Extend the form with first name and last name if it is IBAN payment pointer

Create account (verification of payee) on GateHub side

When creating incoming payment, pass the vopNonce field to metadata

GateHub API request:

This is the URL:

GET https://api.gatehub.net/core/v1/users/:userUuid/accounts

{
    iban: string
    legal_name: string
    recipient_type: "BUSINESS" | "PERSONAL"
    network: 30
}
where userUuid is the userUuid of your organization.
legal name is firstName + lastName combined.

{
    ...
    vop: {
          description: string
          match: "MATCH" | "CLOSE_MATCH" | "NO_MATCH" | "OTHER"
          nonce: string
      }
}
So, from the response, you need to get nonce, and that nonce needs to be passed in the metadata of the [createReceiver mutation](https://rafiki.dev/apis/graphql/backend/#mutation-createReceiver) as a vopNonce field.

Tymmmy:
### Context

- fixes #2143


packages/wallet/backend/src/gatehub/client.ts
```diff
@@ -743,6 +744,28 @@ export class GateHubClient {
...
+  ): Promise<SepaTransactionResponse> {
+    const organizationUuid = this.env.GATEHUB_ORG_ID
+    const url = `${this.apiUrl}/core/v1/users/${organizationUuid}/accounts`
+    const payload = {
+      iban: iban,
+      legal_name: legalName.trim(),
+      recipient_type: 'PERSONAL',
+      network: 30
+    }
+    const response = await this.request<SepaTransactionResponse>(
```
dragosp1011: is this endpoint returning SEPA Details that need to be used in transaction?
or creating a transaction?

packages/wallet/frontend/src/pages/send.tsx
```diff
@@ -267,14 +286,33 @@ const SendPage: NextPageWithLayout<SendProps> = ({ accounts }) => {
...
+              receiver: data.receiver,
+              legalName: data.legalName || ''
+            })
+            if (responseSEPA.success && responseSEPA.result) {
+              sepaDetails.nonce = responseSEPA.result.vop.nonce
+              sepaDetails.match = responseSEPA.result.vop.match
+              sepaDetails.description = responseSEPA.result.vop.description
+            }
+
+            await sleep(10000)
```
dragosp1011: why is this needed?

dragosp1011 (approved):

PR #2188, opened 2026-03-24
refactor: support for mockgatehub in local environment

bosbaber:
Introduces various changes to the packages to allow for a local environment where we can easily change between pointing to the real Gatehub Sandbox environment, or to a locally hosted mock instance of MockGatehub.

This is part of the greater effort of implementing real Continuous Integration for this project to allow for long term maintenance.


packages/wallet/frontend/src/utils/helpers.ts
```diff
@@ -44,7 +44,7 @@ export const formatAmount = (args: FormatAmountArgs): FormattedAmount => {
 
   const scaledValue = Number(`${value}e-${assetScale}`)
   const flooredValue =
-    Math.floor(Math.round(scaledValue * 10 ** displayScale)) /
+    Math.floor(Math.round(scaledValue * 10 ** displayScale) / 10 ** 4) /
```
dragosp1011: is this change needed?

bosbaber: Good point. Thank you.
`formatAmount` is now the same as that of main.

package.json
```diff
@@ -22,17 +22,14 @@
     "format": "pnpm prettier:write && pnpm lint:fix",
     "lint:check": "eslint --max-warnings=0 .",
     "lint:fix": "eslint --max-warnings=0 --fix .",
-    "compose": "docker compose -f ./docker/dev/docker-compose.yml",
-    "compose:prod": "docker compose -f ./docker/prod/docker-compose.yml",
+    "compose": "docker compose -f ./local/docker-compose.yml",
```
dragosp1011: as FE were added to dev compose files, the dev commands need no be updated

pnpm-lock.yaml
dragosp1011, line 1: seems to not be in sync with current branch

local/.env.example
```diff
@@ -0,0 +1,116 @@
...
+# GATEHUB_SETTLEMENT_WALLET_ADDRESS=<your-sandbox-settlement-address>
+# GATEHUB_ORG_ID=<your-sandbox-org-id>
+# GATEHUB_CARD_APP_ID=<your-sandbox-card-app-id>
+
+# Wallet backend rate limits and product codes (optional)
+RATE_LIMIT=false
+RATE_LIMIT_LEVEL=LAX
+GATEHUB_ACCOUNT_PRODUCT_CODE=DEFAULT
+GATEHUB_CARD_PRODUCT_CODE=DEFAULT
+GATEHUB_NAME_ON_CARD=TEST USER
```
dragosp1011: `GATEHUB_NAME_ON_CARD: [ 'Only alphanumeric characters are allowed' ]`, change to `TESTUSER` maybe

local/docker-compose.yml
```diff
@@ -0,0 +1,287 @@
...
+      AUTH_DOMAIN: ${WALLET_BACKEND_AUTH_DOMAIN:-http://rafiki-auth:3006}
+      AUTH_IDENTITY_SERVER_SECRET: ${AUTH_IDENTITY_SERVER_SECRET:-auth-secret-key-12345}
+      RAFIKI_WEBHOOK_SIGNATURE_SECRET: ${RAFIKI_SIGNATURE_SECRET:-327132b5-99e9-4eb8-8a25-2b7d7738ece1}
+      SENDGRID_API_KEY: ${SENDGRID_API_KEY:-}
+      FROM_EMAIL: ${FROM_EMAIL:-noreply@testnet.local}
+      SEND_EMAIL: ${SEND_EMAIL:-false}
+      REDIS_URL: ${REDIS_URL:-redis://redis:6379/0}
+      GATEHUB_API_BASE_URL: ${GATEHUB_API_BASE_URL:-http://mockgatehub:8080}
+      GATEHUB_ENV: ${GATEHUB_ENV:-sandbox}
+      GATEHUB_IFRAME_BASE_URL: ${GATEHUB_IFRAME_BASE_URL:-http://localhost:8080}
```
dragosp1011: the env var is defined but it is not used in the app, resulting in incorrect iframe link on FE `http://mockgatehub:8080/?bearer=iframe-token-4dfcdb88-fe38-4906-9b34-fb3a832a3e30`

local/docker-compose.yml
```diff
@@ -0,0 +1,287 @@
...
+      - '15434:5432'
+    restart: unless-stopped
+    networks:
+      - testnet
+    volumes:
+      - pg-data:/var/lib/postgresql/data
+      - ./init/dbinit.sql:/docker-entrypoint-initdb.d/init.sql
+
+  # MockGatehub - Mock Gatehub API service for local development
+  mockgatehub:
```
dragosp1011: on manual deposits the webhook has `"deposit_type":"external"` , it should have a different value - this is causing double entries in trx list now

dragosp1011: another issue: on trx, the account balance show double of the amount in the trx - but the gain should actually be 0 as the trx is between 2 WA in the same account

dragosp1011: looks like the balance is increased even when it should be decreased in a trx

Tymmmy:
Looks like for Safari the page does not go to KYC, it gets stuck in Login (with chrome it gets to the KYC with problem)

https://github.com/user-attachments/assets/9c096ab0-9b06-41ba-8303-a82ee0034db5

Tymmmy:
KYC displays: 
image

local/docker-compose.yml
bosbaber: So it turns out that Mockgatehub was doing an additional `pending` webhook which caused the double transaction issue. I checked and this is out of spec and I'm not sure why that ended up in mockgatehub.

I removed it from mockgatehub in 1.12.4 and should be resolved.

```mermaid
sequenceDiagram
    actor User
    participant FE as Wallet Frontend
    participant WBE as Wallet Backend
    participant MGH as MockGatehub

    User->>FE: Clicks Deposit on account page
    FE->>User: Opens DepositDialog popup

    User->>FE: Fills amount, clicks Submit
    FE->>WBE: POST /accounts/:accountId/fund<br/>{amount: 11}

    WBE->>MGH: POST /core/v1/transactions<br/>{type:1(DEPOSIT), network:8(MANUAL),<br/>receiving_address, vault_uuid, amount}
    MGH->>MGH: Creates transaction (status: PENDING)

    MGH-->>WBE: Webhook: core.deposit.completed<br/>{deposit_type:"external", status:"pending", tx_uuid}
    MGH->>WBE: 201 Created {transaction}
    WBE->>WBE: handleDepositWebhook()<br/>→ INSERT Transaction row #1 into DB

    WBE->>FE: 200 OK "Account funded"
    FE->>User: Shows "Deposit success" toast

    Note over MGH: 2 second delay

    MGH-->>WBE: Webhook: core.deposit.completed<br/>{deposit_type:"external", status:"completed", tx_uuid}
    MGH->>MGH: Updates balance in MockGatehub store
    WBE->>WBE: handleDepositWebhook()<br/>→ INSERT Transaction row #2 into DB ⚠️ BUG

    User->>FE: Navigates to Transactions page
    FE->>User: Shows 2 transaction rows for the same deposit
```

I do worry that this points to other problems inside TestNet which might need some attention. I will create separate tasks for the following hardening issues which we can discuss separately
- only process webhooks where `data.status === 'completed'`
- Before inserting, check whether a `Transaction` row with `paymentId === data.tx_uuid` already exists and skip if so

packages/boutique/backend/src/order/service.ts
```diff
@@ -4,7 +4,7 @@ import { TransactionOrKnex } from 'objection'
 import { Logger } from 'winston'
 import { InternalServerError, NotFound } from '@shared/backend'
 
-interface OrderItemParams extends Pick<OrderItem, 'productId' | 'quantity'> {}
+type OrderItemParams = Pick<OrderItem, 'productId' | 'quantity'>
```
bosbaber, line 7: This change was because of the new v8 linting rules
`no-empty-interface/no-empty-object-type`

local/config/certs/local.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
bosbaber, line 1: We can re-use the same key for the local environments to help keep things consistent. There is no security problem here.

packages/boutique/backend/src/config/env.ts
```diff
@@ -1,16 +1,24 @@
...
-    .default('postgres://postgres:password@localhost:5433/boutique_backend'),
-  PAYMENT_POINTER: z.string().default('replace-me'),
-  KEY_ID: z.string().default('replace-me'),
-  PRIVATE_KEY: z.string().default('replace-me'),
-  REDIS_URL: z.string().default('redis://redis:6379/0')
+  PORT: z.coerce.number(),
+  NODE_ENV: z.string().min(1),
+  FRONTEND_URL: z.string().url(),
+  DATABASE_URL: z.string().url(),
+  PAYMENT_POINTER: nonPlaceholderString,
```
bosbaber: I felt that users will keep getting this wrong. It is just too easy to copy the address with the $ sign instead of https. So now we will fail early.

packages/boutique/backend/src/config/env.ts
```diff
@@ -1,16 +1,24 @@
...
+  KEY_ID: nonPlaceholderString,
+  PRIVATE_KEY: nonPlaceholderString,
```
bosbaber: Added base64 validation here to fail early when someone might have pasted the key that included the newlines somehow.

packages/boutique/backend/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "@boutique/backend",
   "scripts": {
-    "start": "node dist/index.js",
+    "start": "sh -c 'set --; [ -f .env.local ] && set -- \"$@\" --env-file=.env.local; [ -f .env ] && set -- \"$@\" --env-file=.env; node \"$@\" dist/index.js'",
```
bosbaber, line 4: Frist load the `.env.local` and then the `.env` file. This way the user can override.

packages/boutique/frontend/src/app/checkout/components/pay-button.tsx
```diff
@@ -2,8 +2,10 @@ import { PayWithInterledgerMark } from '@/components/icons.tsx'
 import { Button, ButtonProps } from '@/components/ui/button.tsx'
 import { cn } from '@/lib/utils.ts'
 
-export interface PayButtonProps
-  extends Omit<ButtonProps, 'variant' | 'size' | 'aria-label'> {}
+export type PayButtonProps = Omit<
```
bosbaber, line 5: Linter was complaining

packages/boutique/frontend/src/components/navigation.tsx
```diff
@@ -19,7 +19,7 @@ const NAV_LINKS = [
   }
 ]
 
-interface MainNavLinkProps extends Omit<NavLinkProps, 'className'> {}
+type MainNavLinkProps = Omit<NavLinkProps, 'className'>
```
bosbaber, line 22: linting `no-empty-interface/no-empty-object-type family`

packages/wallet/backend/src/gatehub/client.ts
bosbaber, line 1: This file is one of the major changes to this repository. Note that we are not assuming the structure of the Gatehub URLs anymore. They have to be configured.

package.json
bosbaber: After a long back and forth with myself on this topic I decided to make the FE run on the host and not in docker. The reason is because when running FE in docker on a Mac then the watch mechanism does always work as expected. So in the end I reverted back to have the applications run on the host for the local environment.

dragosp1011 (approved):

PR #2189, opened 2026-03-24
test: e2e tests for basic functionality

bosbaber:
- This PR introduces the e2e framework and the first set of e2e tests. This is to establish the framework on which we will be building further tests.
- No changes made to the `packages` folder, this PR should not modify application behaviour at all
- Reviewers should please see if the e2e tests run on their own machines before approving.


e2e/features/steps/cross-currency-transfer.steps.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  // Select the first wallet address option
+  const firstWalletOption = page.locator('[role="option"]').first()
+  await expect(firstWalletOption).toBeVisible()
+  await firstWalletOption.click()
+  await flow.takeScreenshot('wallet-address-selected')
+})
+
+Then(
+  'I should see the recipient address input field',
```
dragosp1011, line 99: looks like this test is just checking the form and it does not do any transfers

dragosp1011 (approved):

PR #2200, opened 2026-06-03
feat: change the Grants page to match the Transactions page

mirunagherman:
### Context

- Linear Issue INT-564 - Change the grants page list to match the one from transactions list

### Changes
- In the Grant Details popup, renamed the "State" label to "Status" and formatted the "Created at" date.
- The Status column is hidden on mobile devices.
- The Date column is now the first column and supports ascending/descending sorting.
- Grants are grouped by day, with the day displayed as a group header and the time shown in each grant row.
- Replaced the previous pagination approach with page-based pagination.
- Page numbers are displayed at the bottom when multiple pages are available.
- Added arrow buttons for pagination on smaller screens.


Tymmmy (changes requested):
I tested the functionality and it works as expected, one thing I noticed, for example on Transactions page if you go to the 4th page (any page that is not the first) and sort by Date, the page will remain in the 4th page. The Grants page, if you go to the 4th page, and you sort by Date, it will skip to the firts page. If you can check this issue out, the rest seems fine and on point.

Tymmmy (dismissed):

packages/wallet/backend/src/grant/validation.ts
```diff
@@ -1,7 +1,19 @@
...
 export const grantResponseSchema = z.object({
   body: z.object({
     response: z.enum(['accept', 'reject'])
   })
 })
+
+export const grantListRequestSchema = z.object({
+  body: z.object({
+    page: z.coerce.number().int().nonnegative().default(0),
+    pageSize: z.coerce.number().int().positive().default(10),
```
adrianboros: let's also bound the size of the pageSize also, lets say to .max(100)

packages/wallet/shared/src/types/grant.ts
```diff
@@ -51,6 +51,7 @@ type GrantsPageInfo = {
   startCursor: string
```
adrianboros: if we don't use the `endCursor` and `startCursor` anymore maybe we could remove them

packages/wallet/frontend/src/pages/grants/index.tsx
```diff
@@ -59,48 +81,89 @@ const GrantsPage: NextPageWithLayout<GrantsPageProps> = () => {
...
+                    groupByDate !==
+                    formatDateNoTime({ date: grant.node.createdAt })
+                  ) {
+                    groupByDate = formatDateNoTime({
+                      date: grant.node.createdAt
+                    })
+                    showDateHeader = true
+                  }
+                  return (
+                    <>
```
adrianboros: Move the key prop from the inner <Table.Row> to the top-level empty fragment <>...</> inside grants.edges.map() to fix React reconciliation warnings.
Example:
`return (
  <Fragment key={grant.node.id}>
    ...
  </Fragment>
)`

adrianboros (approved):

PR #2204, opened 2026-06-08
fix: will now better log for login problems

bosbaber:
At the moment we are only having deafening silence when login fails.

This PR aims to change that.

### DIFF
--- a/packages/wallet/backend/src/auth/service.ts
+++ b/packages/wallet/backend/src/auth/service.ts
@@ -30,6 +30,20 @@ interface IAuthService {
   signUp(args: SignUpArgs): Promise<User>
 }
 
+const maskEmail = (email: string): string => {
+  const [localPart, domain] = email.split('@')
+
+  if (!localPart || !domain) {
+    return '***'
+  }
+
+  if (localPart.length <= 2) {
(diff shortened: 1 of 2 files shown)


packages/wallet/backend/src/auth/service.ts
```diff
@@ -95,15 +109,24 @@ export class AuthService implements IAuthService {
...
+      this.logger.info(
+        `Login rejected: no account found for ${maskEmail(args.email)}`
+      )
       throw new Unauthorized('Invalid credentials.')
     }
 
     const isValid = await user.verifyPassword(args.password)
     if (!isValid) {
+      this.logger.info(
+        `Login rejected: invalid password for ${maskEmail(args.email)}`
```
dragosp1011, line 121: maybe use invalid credentials here as well

dragosp1011 (approved):

packages/wallet/backend/src/auth/service.ts
bosbaber: do youmean we should log the invalid credentials used?

dragosp1011: no, just use the same wording as the error

PR #2214, opened 2026-06-22
feat: add merchant onboarding field definitions API

mirunagherman:
### Context

- fixes [INT1-658](https://linear.app/interledger/issue/INT1-658/return-ui-fields-for-linking-merchant-to-the-pos-device)

### Changes
- Add terminal API for retrieving merchant onboarding field definitions
- Add `field_definitions` and `options` tables to store form configuration
- Add terminal model, controller, and service
- Return part of onboarding form fields


packages/wallet/backend/src/terminal/controller.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  getOnboardingFormDefinition = async (
+    _req: Request,
+    res: Response,
+    next: NextFunction
+  ) => {
+    try {
+      const formDefinition =
+        await this.terminalService.getOnboardingFormDefinition()
+      res.status(200).json(formDefinition)
```
DaevMithran:
```suggestion
      res.status(200).json(toSuccessResponse(formDefinition))
```

packages/wallet/backend/src/app.ts
```diff
@@ -454,6 +457,12 @@ export class App {
       interledgerCardController.terminate
     )
 
+    // Terminal
+    router.get(
+      '/terminals/onboarding',
```
DaevMithran, line 462: Non-blocking: 

Should this endpoint be public? or protected by a custom terminal auth middleware or a rate limiter if it's left public.

DaevMithran (changes requested):

packages/wallet/backend/src/app.ts
golobitch: Agreed with Daev. This endpoint should not be public. It is called only from Merchant API - We can pass some API key (uuid) in the request head for example and you can check if the header is correct - need to discuss about this.

Does testnet wallet has proper HTTP signatures (RFC 9421)?

beniaminmunteanu: test wallet has different signature schemas based on who it's talking to.

In regards to RFC 9421 -> That is done by Rafiki, so the test wallet backend doesn't need to implement it directly. It only handles the key management. and [verifies the webhooks](https://rafiki.dev/integration/requirements/webhook-events/#verify-webhook-signatures) with custom HMAC-SHA256.
The same HMAC signing is done for outbound requests towards the rafiki graphql api.

For gatehub, it also does signature verification, HMAC-SHA256  and outbound signing as well in the same style.

And stripe is handled by the stripe sdk.

So afaik the test wallet backend itself does not implement the RFC for http signatures, in case you were thinking of reusing that in some way. @golobitch

beniaminmunteanu: The endpoint can be public here, and we will only make it accessible from the merchant api from a cluster networking perspective.

resolving.

beniaminmunteanu (commented):
Please add unit tests for the code additions.

DaevMithran (approved):

DaevMithran:
LGTM!

PR #2235, opened 2026-09-01
feat: introduce prom client metrics

bosbaber:
### Context
We have a problem where are really struggle to understand the reason of performance bottlenecks. This is especially true whenever the wallet appears unexpectedly slow.

This pull request adds Prometheus-compatible HTTP metrics to both the Boutique and Wallet backend services, enabling improved observability and monitoring. It introduces a new metrics server running on a dedicated port, exposes metrics via a `/metrics` endpoint, and integrates middleware to collect HTTP request metrics. The changes are thoroughly tested and configurable via Helm charts and environment variables.


bosbaber:
I'm closing this PR because we will introduce this in v1 instead

